PR #33, opened 2025-01-09
the components should only be destoyed if explitly destroyed by caller

TrevorDev:
https://github.com/spatial-web/XRSDK/issues/32

### DIFF
--- a/builder/visionOSApp/web-spatial/libs/SpatialEntity.swift
+++ b/builder/visionOSApp/web-spatial/libs/SpatialEntity.swift
@@ -135,8 +135,6 @@ class SpatialEntity: SpatialObject {
             wg.removeEntity(self)
         }
 
-        // handle components destroy
-        components.forEach { $0.destroy() }
         components = []
 
         setParent(parentEnt: nil)


builder/visionOSApp/web-spatial/libs/SpatialEntity.swift
```diff
@@ -135,8 +135,6 @@ class SpatialEntity: SpatialObject {
             wg.removeEntity(self)
         }
 
-        // handle components destroy
-        components.forEach { $0.destroy() }
```
TrevorDev, line 139: @walton007 does this seem right to you? To me I expect to be able to create a component, destroy an entity with that component and then add it to another ent

walton007: For my original design,   an component should belong to an entity.  It's part of an entity. Because entity can have enable/disable state, which will influence to component's active/inAcitive state.  So when an entity is destroyed, all of entities'components should be destroyed.

walton007: So component can be only created by  entity.createComponent().  Component can destroyed by component.destroy()

walton007: Webview can be a shared resource, which can be allocated to a SpatialWindowComponent. When a SpatialWindowComponent is destroyed, it can choose recycle that Webview to WebviewManager, or decide to destroy that Webview resource by indicate to WebviewManager

walton007: From my previous experience, components should not be shared, but assets like material/texture/image/webview can be shared

PR #53, opened 2025-01-15
[Review] Jsapi review pass

TrevorDev:
Updated comments and added TODO tag to APIs we should review together.

### DIFF
--- a/runtime/src/core/component/SpatialComponent.ts
+++ b/runtime/src/core/component/SpatialComponent.ts
@@ -1,7 +1,14 @@
 import { WebSpatial, WebSpatialResource } from '../private/WebSpatial'
 import { SpatialEntity } from '../SpatialEntity'
 import { SpatialObject } from '../SpatialObject'
+
+/** @hidden */
 export class SpatialComponent extends SpatialObject {
+  /**
+   * Gets the entity this component is attached to
+   * [TODO] should this be removed?
+   * @returns entity or null
+   */
   async getEntity() {
     let reqResp: any = await WebSpatial.updateResource(
       WebSpatial.getCurrentWebPanel(),
--- a/runtime/src/core/component/SpatialInputComponent.ts
+++ b/runtime/src/core/component/SpatialInputComponent.ts
@@ -1,6 +1,10 @@
 import { Vec3 } from '../private/WebSpatial'
 import { SpatialComponent } from './SpatialComponent'
 
+/**
+ * Translate event, matching similar behavior to https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/drag_event
+ * [TODO] need to support onclick event
+ */
 type TranslateEvent = {
   eventType: 'dragstart' | 'dragend' | 'drag'
   translate?: Vec3
@@ -9,8 +13,14 @@ type TranslateEvent = {
  * Used to handle input events on an entity
  */
 export class SpatialInputComponent extends SpatialComponent {
+  /** @hidden */
   public _gotEvent(data: any) {
     this.onTranslate(data)
   }
+
+  /**
+   * Callback fired when a translate event occurs
+   * @param data translate event data
+   */
   public onTranslate(data: TranslateEvent) {}
 }
--- a/runtime/src/core/component/SpatialWindowComponent.ts
+++ b/runtime/src/core/component/SpatialWindowComponent.ts
@@ -49,6 +49,7 @@ export class SpatialWindowComponent extends SpatialComponent {
   /**
    * Sets the resolution of the window, the resulting dimensions when rendered will be equal to 1/1360 units
    * eg. if the resolution is set to 1360x1360 it will be a 1x1 plane
+   * [TODO] should better document how this translates to dimensions when in app sapce, 1360 doesnt seem to be in any docs
    * @param x width in pixels
    * @param y height in pixels
    */
@@ -58,6 +59,10 @@ export class SpatialWindowComponent extends SpatialComponent {
     })
   }
 
+  /**
+   * [TODO] should this be on entity instead?
+   * @param rotationAnchor
+   */
   async setRotationAnchor(rotationAnchor: Vec3) {
     await WebSpatial.updateResource(this._resource, {
       rotationAnchor: rotationAnchor,
--- a/runtime/src/core/resource/SpatialMesh.ts
+++ b/runtime/src/core/resource/SpatialMesh.ts
@@ -1,3 +1,7 @@
 import { SpatialObject } from '../SpatialObject'
 
+/**
+ * Mesh asset containing geometry
+ * [TODO] should the naming of resources end in Resource like component?
+ */
 export class SpatialMesh extends SpatialObject {}
--- a/runtime/src/core/component/SpatialModelComponent.ts
+++ b/runtime/src/core/component/SpatialModelComponent.ts
@@ -32,6 +32,7 @@ export class SpatialModelComponent extends SpatialComponent {
     })
   }
 
+  /** @hidden */
   async _syncMaterials() {
     await this.setMaterials(this.cachedMaterials)
   }
--- a/runtime/src/core/component/SpatialViewComponent.ts
+++ b/runtime/src/core/component/SpatialViewComponent.ts
@@ -6,6 +6,7 @@ import { SpatialComponent } from './SpatialComponent'
  * Child entities will be added within this volume's space
  * Defaults to having 1x1x1 meter dimensions
  * Resolution defaults to 100x100 pixels
+ * Only will be displayed on entities in "ROOT" or "DOM" space
  */
 export class SpatialViewComponent extends SpatialComponent {
   /**
--- a/runtime/src/core/resource/SpatialPhysicallyBasedMaterial.ts
+++ b/runtime/src/core/resource/SpatialPhysicallyBasedMaterial.ts
@@ -6,8 +6,17 @@ import { SpatialObject } from '../SpatialObject'
  * PBR material which can be set on a SpatialModelComponent
  */
 export class SpatialPhysicallyBasedMaterial extends SpatialObject {
+  /**
+   * Base color of the material containing rgba between 0 and 1
+   */
   baseColor = { r: 0.0, g: 0.7, b: 0.7, a: 1.0 }
+  /**
+   * PBR metalic value between 0 and 1
+   */
   metallic = { value: 0.5 }
+  /**
+   * PBR roughness value between 0 and 1
+   */
   roughness = { value: 0.5 }
 
   _modelComponentAttachedTo: { [key: string]: SpatialModelComponent } = {}


runtime/src/core/SpatialEntity.ts
```diff
@@ -77,10 +88,22 @@ export class SpatialEntity extends SpatialObject {
     })
   }
 
+  /**
+   * Sets the coordinate space of this entity (Default: App)
+   * "App" = game engine style coordinates in meters
+   * "Dom" = Windowing coordinates in dom units (eg. 0,0,0 is top left of window)
+   * "Root" = Coordinate space is ignored and content is displayed and updated as windowGroup's root object, window groups can only have one root entity
+   * [TODO] review this api
```
walton007, line 96: this is a very complex api. Which coordinate space should be used in which scenario? @dexteryy  do you understand the difference of these coordinate space?

runtime/src/core/Spatial.ts
```diff
@@ -4,6 +4,11 @@ import { SpatialSession } from './SpatialSession'
  * Base object designed to be placed on navigator.spatial to mirror navigator.xr for webxr
  */
 export class Spatial {
+  /**
+   * Requests a session object from the browser
+   * @returns The session or null if not availible in the current browser
+   * [TODO] discuss implications of this not being async
```
walton007, line 10: if it could be a sync api, why use async?

runtime/src/core/SpatialEntity.ts
```diff
@@ -35,12 +39,16 @@ export class SpatialEntity extends SpatialObject {
 
   /**
    * Attaches a component to the entity to be displayed
+   * [TODO] review pass by value vs ref and ownership model for this
```
walton007, line 42: a component should belong to only one entity. If a component have already attached to an entity, it should be be added to other entities. So some component owner check can be added here to avoid duplicate adding. 
If a component have already been added to an entity, the entity have the ownership the component

runtime/src/core/SpatialEntity.ts
```diff
@@ -77,10 +88,22 @@ export class SpatialEntity extends SpatialObject {
...
+   * @param space coordinate space mode
+   */
   async setCoordinateSpace(space: 'App' | 'Dom' | 'Root') {
     await WebSpatial.updateResource(this._entity, { setCoordinateSpace: space })
   }
 
+  /**
+   * Query the 3d boudning box of the entity
+   * @returns The bounding box of the entity
+   */
```
walton007, line 106: if an entity have a SpatialWindowComponent only, what's the meaning of 3d bounding box of the entity?
Which decides the bbox of an entity?

runtime/src/core/SpatialSession.ts
```diff
@@ -11,26 +11,24 @@ import {
...
-  /** @hidden */
-  _currentFrame = new SpatialFrame()
   /** @hidden */
   _animationFrameCallbacks = Array<animCallback>()
   /** @hidden */
   _frameLoopStarted = false
 
   /**
    * Request a callback to be called before the next render update
+   * [TODO] might want to rework this to be triggered by native code or better handle frame timing
```
walton007, line 31: which native code will trigger this raf ?

runtime/src/core/SpatialSession.ts
```diff
@@ -62,6 +61,7 @@ export class SpatialSession {
 
   /**
    * Creates a WindowComponent
+   * [TODO] should creation of components be moved to entity? and these made private?
```
walton007, line 64: if a component should only exist in one entity, then entity can create and add a specific component. Devs don't need to create component from session, this api can be moved to entity

runtime/src/core/SpatialSession.ts
```diff
@@ -183,6 +185,10 @@ export class SpatialSession {
     }
   }
 
+  /**
+   * [TODO] should these log apis be private?
```
walton007, line 189: do we still need any log related apis?

runtime/src/core/SpatialSession.ts
```diff
@@ -158,6 +159,7 @@ export class SpatialSession {
   /**
    * Retrieves the window for this page
    * @returns the window component corresponding to the js running on this page
+   * [TODO] discuss implications of this not being async
```
walton007, line 162: why make it async?

runtime/src/core/SpatialSession.ts
```diff
@@ -213,23 +219,31 @@ export class SpatialSession {
...
    * Debugging only, used to ping the native renderer
+   * [TODO] make this private
    */
   async ping(msg: string) {
     return await WebSpatial.ping(msg)
   }
 
   /**
    * Debugging to get internal state from native code
+   * [TODO] make this private
```
walton007, line 230: agree

runtime/src/core/SpatialSession.ts
```diff
@@ -213,23 +219,31 @@ export class SpatialSession {
...
    * @returns data as a js object
    */
   async getStats() {
     return (await WebSpatial.getStats()) as any
   }
 
+  /**
+   * [TODO] make this private
```
walton007, line 238: agree

runtime/src/core/SpatialSession.ts
```diff
@@ -213,23 +219,31 @@ export class SpatialSession {
...
+   */
   async inspect(spatialObjectId: string = WebSpatial.getCurrentWebPanel().id) {
     return WebSpatial.inspect(spatialObjectId)
   }
 
+  /**
+   * [TODO] make this private
```
walton007, line 245: agree

runtime/src/core/SpatialSession.ts
```diff
@@ -259,6 +276,12 @@ export class SpatialSession {
 
   // Retreives the window group that is the parent to this spatial web page
   private static _currentWindowGroup = null as null | SpatialWindowGroup
+
+  /**
+   * Gets the current window group for the window
+   * [TODO] discuss what happens if it doesnt yet have a windowgroup
```
walton007, line 282: how is it possible that current window group not exist?

runtime/src/core/SpatialTransform.ts
```diff
@@ -1,5 +1,6 @@
 /**
  * Transform containing position, orientation and scale
+ * [TODO] is DOMPoint the right type to use?
```
walton007, line 3: I don't like DOMPoint, because it's not related to DOM. I prefer Vec3/Vec4, which is more common in web3d

runtime/src/core/component/SpatialComponent.ts
```diff
@@ -1,7 +1,14 @@
 import { WebSpatial, WebSpatialResource } from '../private/WebSpatial'
 import { SpatialEntity } from '../SpatialEntity'
 import { SpatialObject } from '../SpatialObject'
+
+/** @hidden */
 export class SpatialComponent extends SpatialObject {
+  /**
+   * Gets the entity this component is attached to
+   * [TODO] should this be removed?
```
walton007, line 9: why remove this getEntity method? I think it's useful

runtime/src/core/component/SpatialInputComponent.ts
```diff
@@ -1,6 +1,10 @@
 import { Vec3 } from '../private/WebSpatial'
 import { SpatialComponent } from './SpatialComponent'
 
+/**
+ * Translate event, matching similar behavior to https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/drag_event
+ * [TODO] need to support onclick event
```
walton007, line 6: more work should be done related to interaction. We can improve the interactive event api through model interaction @dexteryy

runtime/src/core/component/SpatialWindowComponent.ts
```diff
@@ -58,6 +59,10 @@ export class SpatialWindowComponent extends SpatialComponent {
     })
   }
 
+  /**
+   * [TODO] should this be on entity instead?
```
walton007, line 63: the anchor property is currently consumed by SpatialWebviewUI only, this is the reason I add it as property of SpatialWindowComponent. 
If the anchor property is moved to Entity, what's the meaning if SpatialWindowComponent is not attached to the entity?

runtime/src/core/resource/SpatialMesh.ts
```diff
@@ -1,3 +1,7 @@
 import { SpatialObject } from '../SpatialObject'
 
+/**
+ * Mesh asset containing geometry
+ * [TODO] should the naming of resources end in Resource like component?
```
walton007, line 5: Agree

runtime/src/core/component/SpatialWindowComponent.ts
```diff
@@ -49,6 +49,7 @@ export class SpatialWindowComponent extends SpatialComponent {
   /**
    * Sets the resolution of the window, the resulting dimensions when rendered will be equal to 1/1360 units
    * eg. if the resolution is set to 1360x1360 it will be a 1x1 plane
+   * [TODO] should better document how this translates to dimensions when in app sapce, 1360 doesnt seem to be in any docs
```
walton007, line 52: I asked GPT about the pointerPerMeter. It seems that value of content.pointerPerMeter is determined by several factors. What's the behavior in any other platform( like AndroidXR)?

runtime/src/core/Spatial.ts
TrevorDev: The reason is it might not always be support in a sync way on all systems. In the future we might want to ask over jsb if its supported. There are some apis like requestSession and getCurrentWindowGroup that potentially might have some quirks due to them being sync currently. One of the reasons they are nice to be sync is because they can allow us to set the style of the page before the page is displayed to the user

runtime/src/core/SpatialEntity.ts
TrevorDev: ok, i think this seems reasonable I can do that

runtime/src/core/SpatialEntity.ts
TrevorDev: maybe we can discuss this in the next meeting

runtime/src/core/SpatialEntity.ts
TrevorDev: it might be good to return the dimensions the entity(with window) would have if displayed in a volume. I think right now its 1360px = 1 meter

runtime/src/core/SpatialSession.ts
TrevorDev: realitykit system api has an onRender api, worth considering using that to trigger so we done render to slow/fast

runtime/src/core/SpatialSession.ts
TrevorDev: ok, I can do that

runtime/src/core/SpatialSession.ts
TrevorDev: when we call this, it sets the internal object as "current" so the js api doesn't acually know 100% that it exists (although it likely does due to this script running within it)

runtime/src/core/SpatialSession.ts
TrevorDev: I think we should remove them and make them private (maybe just have a debug log api)

runtime/src/core/SpatialSession.ts
TrevorDev: if you create a windowComponent, it can be offscreen I think, so if its not added to a windowGroup right away, this will return a "current" reference, but internally it doesn't exist.

runtime/src/core/SpatialTransform.ts
TrevorDev: Yea, I think that sounds reasonable, the reason I chose domPoint was because webXR uses that, maybe we can use our own vec3 instead for now.

runtime/src/core/component/SpatialComponent.ts
TrevorDev: If the component is always attached to an entity, the developer should already have a reference to the entity right?

runtime/src/core/component/SpatialWindowComponent.ts
TrevorDev: yea, this api is very confusing on AVP and not documented too well. Will need to dig deeper, potentially user will have to call getBoundingBox to know the size?

runtime/src/core/component/SpatialWindowComponent.ts
TrevorDev: it would also potentially need to work with model on an entity.

PR #55, opened 2025-01-15
change createWindowContext to use webspatial://createWindowContext

### LINKED ISSUE #52: createWindowContext should using custom url
As developer may use `window.open()`

we should not use it  in our `SpatialSession.createWindowContext()`

Maybe use `webspatial://createWindowContext`

mzvast:
this will close #52

### DIFF
--- a/runtime/src/core/SpatialSession.ts
+++ b/runtime/src/core/SpatialSession.ts
@@ -279,7 +279,7 @@ export class SpatialSession {
 
   // Creates a window context object that is compatable with SpatialWindowComponent's setFromWindow API
   async createWindowContext() {
-    let openedWindow = window.open()
+    let openedWindow = window.open('webspatial://createWindowContext')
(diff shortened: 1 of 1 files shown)


runtime/src/core/SpatialSession.ts
```diff
@@ -279,7 +279,7 @@ export class SpatialSession {
 
   // Creates a window context object that is compatable with SpatialWindowComponent's setFromWindow API
   async createWindowContext() {
-    let openedWindow = window.open()
+    let openedWindow = window.open('webspatial://createWindowContext')
```
TrevorDev, line 282: Do we need any other native side changes to support this or is that already done?

mzvast: It works with no change to other side, except some warnings about the URL like`Navigation failed!!! unsupported URL`. It can be supressed later on.

PR #66, opened 2025-01-17
Model support desktop mode via webgl

TrevorDev:
https://github.com/spatial-web/XRSDK/issues/65

### DIFF
--- a/react/src/spatial-react-components/Model.tsx
+++ b/react/src/spatial-react-components/Model.tsx
@@ -60,12 +60,14 @@ export const Model = forwardRef((props: ModelProps, ref: ModelRef) => {
   let session = getSession()
   if (!session) {
     useEffect(() => {
-      React.Children.forEach(props.children, async element => {
+      React.Children.toArray(props.children).some((element: any) => {
         var src = element.props.src as string
         var fileExt = src.split('.').pop()
         if (fileExt == 'glb' || fileExt == 'gltf') {
           setGlbSrc(src)
+          return true // exit early
         }
+        return false
       })
     }, [])
     return (


react/src/spatial-react-components/Model.tsx
```diff
@@ -43,13 +52,31 @@ export type ModelRef = Ref<{
...
+  var [glbSrc, setGlbSrc] = useState('')
   initializeSpatialOffset(props.spatialOffset!)
 
   const opacity = props.opacity ?? 1
 
   let session = getSession()
   if (!session) {
-    return <div className={props.className}> model can display in AVP only</div>
+    useEffect(() => {
+      React.Children.forEach(props.children, async element => {
```
mzvast: This loop will not exit when see the first matched one, so it will select the last one. Is this expected?

testServer/index.tsx
```diff
@@ -275,22 +275,18 @@ function App() {
...
-            </div>
-          ) : (
-            <div className="w-full h-52">
-              <div className="w-full h-full bg-white bg-opacity-25 rounded-xl">
-                Model goes here
-              </div>
-            </div>
-          )}
+          <div className="w-full h-52">
+            <Model className="w-full h-full bg-white bg-opacity-25 rounded-xl">
```
mzvast, line 279: Will the Model element itself have a src property? As this [link](https://immersive-web.github.io/model-element/#adding-a-model-to-a-document) shows that `<model> source selection follows <video> selection, where the top-level src attribute takes precedence if present, followed by the first compatible <source> element encountered.`

TrevorDev: I think @dexteryy should clarify this for us.

react/src/spatial-react-components/Model.tsx
TrevorDev: Fixed, thanks

PR #67, opened 2025-01-20
Spatial open API

mzvast:
```js
// this will work how users will expect (works like creating a new tab but will createa. new windowGroup)
var a = window.open("url", target) 
a.close()


var config = {defaultSize:{width:900,height:900},resizability:"automatic"}
// Customization of the windowGroup i think we should have new api
var s = await session.spatialOpen("url", {config})
```

### DIFF
--- a/builder/visionOSApp/web-spatial/libs/Utils/WindowGroupMgr.swift
+++ b/builder/visionOSApp/web-spatial/libs/Utils/WindowGroupMgr.swift
@@ -39,8 +39,8 @@ struct WindowGroupPlainDefaultValues {
 extension WindowGroupPlainDefaultValues {
     init(_ options: WindowGroupOptions) {
         defaultSize = CGSize(
-            width: defaultSize?.width ?? DefaultPlainWindowGroupSize.width,
-            height: defaultSize?.height ?? DefaultPlainWindowGroupSize.height
+            width: options.defaultSize?.width ?? DefaultPlainWindowGroupSize.width,
+            height: options.defaultSize?.height ?? DefaultPlainWindowGroupSize.height
         )
         windowResizability = getWindowResizability(options.resizability)
     }


builder/visionOSApp/web-spatial/libs/Utils/WindowGroupMgr.swift
```diff
@@ -77,9 +88,8 @@ class WindowGroupMgr: ObservableObject {
     }
 
     func setToMainSceneCfg() {
-        if let cfg = WindowGroupPlainDefaultValues(pwaConfig.mainScene) {
```
TrevorDev, line 80: why is the if statement no longer needed?

builder/visionOSApp/web-spatial/libs/Utils/CommandManager.swift
```diff
@@ -501,126 +516,6 @@ class CommandDataManager {
         target.completeEvent(requestID: requestID)
     }
 
-    public func handleScene(target: SpatialWindowComponent, requestID: Int, data: JSData) {
```
TrevorDev, line 504: Yay, thanks for figuring out how to remove this :)

runtime/src/core/component/SpatialWindowComponent.ts
```diff
@@ -74,6 +74,17 @@ export class SpatialWindowComponent extends SpatialComponent {
     })
   }
 
+  /**
+   *
+   * Sets webview close behavior
+   * @param {boolean} b
+   */
+  async setIsRoot(b: boolean) {
```
TrevorDev, line 82: Maybe we need another name here? We already have entity.CoordinateSpace "ROOT". Do you have any suggestions? Also the comment should describe what the close behavior will be when set to true or false. Can we also not call the variable b?

PR #85, opened 2025-01-22
implement Model3D

walton007:
basic function of Model3D, which supply feature of Native Model3D


react/src/spatial-react-components/Model3D/Model3D.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  console.log('dbg style', style, layoutDomStyle)
+
+  return (
+    <div className={className} style={layoutDomStyle} ref={layoutInstanceRef} />
+  )
+}
+
+export const Model3D = forwardRef(Model3DComponent)
```
TrevorDev, line 72: Is @dexteryy aligned with this naming, we should remove or make <Model> private if we go with model3D

react/src/spatial-react-components/Model3D/useDetectLayoutDomUpdated.ts
```diff
@@ -0,0 +1,50 @@
+import { useRef, useLayoutEffect, useEffect } from 'react'
+import { useForceUpdate } from '../hooks/useForceUpdate'
+
+export function useDetectLayoutDomUpdated(onDomUpdated: () => void) {
```
TrevorDev, line 4: Why is this logic only for model3D? Dont we need to detect dom layout changes for other components already too?

react/src/spatial-react-components/Model3D/Model3D.tsx
walton007: the name can be changed later. I just don't like too many code to be pushed at once

react/src/spatial-react-components/Model3D/useDetectLayoutDomUpdated.ts
walton007: The similar logic in spatialdiv is useDetectDomRectChange in StandardInstance.
but they have different communication mechanism. 
- StandardInstance sync dom change to PortalInstance by SpatialReactContext
- Model3D use simple onDomUpdated callback

PR #87, opened 2025-01-22
initScene API

mzvast:
closing #80

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "postinstall": "simple-git-hooks",
     "buildPackages": "cd runtime/ && npm run build && cd .. && cd react/ && npm run build && cd .. && cd jsx-runtime/ && npm run build && cd ..",
     "setup": "pnpm install && npm run buildPackages",
-    "watchNPM": "concurrently \"cd runtime/ && npm run start\" \"cd react/ && npm run start\"",
+    "watchNPM": "concurrently \"cd runtime/ && npm run start\" \"cd react/ && npm run start\" \"cd jsx-runtime/ && npm run start\"",
     "testServer": "cd testServer/ && npm run dev -- --host",
     "dev": "concurrently \"npm run watchNPM\" \"npm run testServer\"",
     "test": "echo \"Error: no test specified\" && exit 1",


builder/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift
```diff
@@ -57,6 +59,9 @@ class SpatialWindowComponent: SpatialComponent {
         return inspectInfo
     }
 
+    // if this is root
```
TrevorDev: Maybe add a description of what being root means

builder/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift
```diff
@@ -154,12 +164,24 @@ class SpatialWindowComponent: SpatialComponent {
...
     }
 
     func initFromURL(url: URL) {
         webViewNative = WebViewNative(url: url)
         webViewNative?.webViewRef = self
         _ = webViewNative?.createResources()
+        registerForceStyle()
+    }
+
+    func registerForceStyle() {
```
TrevorDev, line 181: Can you add a comment the goal of this method?

builder/visionOSApp/web-spatial/libs/Utils/CommandManager.swift
```diff
@@ -479,6 +474,34 @@ class CommandDataManager {
 
     public func createWindowGroup(target: SpatialWindowComponent, requestID: Int, data: JSData) {
         if let windowStyle: String = data.windowStyle {
+            // windowID exist in SWC
+
+            // TODO: check url scope
+            // if not in scope open in safari
```
TrevorDev: Can you add more detail to this comment, what does it mean to not be in scope?

react/src/XRApp.ts
```diff
@@ -0,0 +1,116 @@
...
+  deinit() {
+    ;(window as any).open = originalOpen
+  }
+  private configMap: Record<string, WindowGroupOptions> = {} // name=>config
+  private getConfig(name?: string) {
+    if (name === undefined || !this.configMap[name]) return undefined
+    return this.configMap[name]
+  }
+
+  // This not work for some reason, use raw JSB instead
```
TrevorDev: What do you mean by this comment? Maybe we should fix this if its broken?

react/src/XRApp.ts
```diff
@@ -0,0 +1,116 @@
...
+    const newWindow = originalOpen(url, target, features)
+    if (url === CONTEXT_WINDOW_URL) return newWindow
+    // if target is special
+    if (target === '_self' || target === '_parent' || target === '_top') {
+      return newWindow
+    }
+    // should open new scene or focus to
+    // make sure _webSpatialID exist
+    let cnt = 2
+    let timer = setInterval(async () => {
```
TrevorDev, line 64: Can you comment why there is the need for the timer logic?

runtime/src/core/SpatialSession.ts
```diff
@@ -151,8 +151,20 @@ export class SpatialSession {
    * Creates a WindowGroup
    * @returns WindowGroup
    */
-  async createWindowGroup(style: WindowStyle = 'Plain') {
-    return new SpatialWindowGroup(await WebSpatial.createWindowGroup(style))
+  async createWindowGroup(
```
TrevorDev, line 184: Can you update the comments with how the config parameters are expected to work?

runtime/src/core/private/WebSpatial.ts
```diff
@@ -122,15 +122,18 @@ export class WebSpatial {
...
+      ...cfg,
+    })
 
     var result = await new Promise((res, rej) => {
       WebSpatial.eventPromises[cmd.requestID] = { res: res, rej: rej }
       WebSpatial.sendCommand(cmd)
     })
     var res = new WindowGroup()
-    res.id = (result as any).data.createdID
+    res.id = (result as any)?.data?.createdID
```
TrevorDev: why would this stuff not be populated? Dont we want to always have an id?

react/src/XRApp.ts
mzvast: I find this line stop the message from sending. 
[`webViewNative!.webViewHolder.appleWebView!.evaluateJavaScript("window.__WebSpatialUnloaded = true")`](https://github.com/spatial-web/XRSDK/blob/932dc22ef41cf2d9562433f73a22ea752146a299/builder/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift#L268)

It seem to fix another bug, stop messaging when disconnected, but now it stop new window

And I think this line should be removed for now.

runtime/src/core/private/WebSpatial.ts
mzvast: reverted

Issue #44, opened 2025-01-13
spatial layout: hook API 


mzvast:
To satisfy configuration from page itself instead of preconfigured, we should support 

### API as following:

```js
// parent page
window.open("sub_page_url")

```

```js
// sub page
window.xrCurrentSceneDefaults = async (defaultConfig)=>{
    const config = await requestDatabase()
    return {
        defaultSize: config.defaultSize,
        resizability: 'contentSize'
    }
}
```

### Expected behavior:

scene opened using settings resovled from `window.xrCurrentSceneDefaults`

PR #88, opened 2025-01-22
scene hook API

mzvast:
rely on #81 
closing #44

### DIFF
--- a/jsx-runtime/src/injectSceneHook.ts
+++ b/jsx-runtime/src/injectSceneHook.ts
@@ -1,14 +1,6 @@
 import { defaultSceneConfig, XRApp } from '@xrsdk/react'
 
 export function injectSceneHook() {
-  const sendMsg = (msg: any) => {
-    try {
-      //@ts-ignore
-      window.webkit.messageHandlers.bridge.postMessage(JSON.stringify(msg))
-    } catch (error) {
-      console.error(error)
-    }
-  }
   if (window.opener) {
     // see this flag, we have done create the root scene
     if ((window as any)._SceneHookOff) return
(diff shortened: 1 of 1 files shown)


testServer/src/scene/hook.tsx
```diff
@@ -0,0 +1,298 @@
+import React, { useEffect, useRef, useState } from 'react'
+import ReactDOM from 'react-dom/client'
+import { Spatial } from '@xrsdk/runtime'
+import { initScene } from '@xrsdk/react'
+;(window as any).xrCurrentSceneDefaults = async (config: any) => {
```
TrevorDev, line 5: needing to do (window as any) is a pretty awkward api, wish we could have something like

import { initScene, setXRCurrentSceneDefaults } from '@xrsdk/react'

instead

jsx-runtime/src/injectSceneHook.ts
```diff
@@ -0,0 +1,58 @@
...
+    // see this flag, we have done create the root scene
+    if ((window as any)._SceneHookOff) return
+    document.addEventListener('DOMContentLoaded', async () => {
+      try {
+        if (typeof (window as any).xrCurrentSceneDefaults === 'function') {
+          const ans = await (window as any).xrCurrentSceneDefaults?.()
+
+          // raw version of await XRApp.getInstance().show((window as any)._webSpatialID, ans)
+
+          sendMsg({
```
TrevorDev: Why cant you use the remoteCommand api like the rest of our APIs?

eg. var cmd = new RemoteCommand('createWindowGroup', { windowStyle: style })

jsx-runtime/src/injectSceneHook.ts
```diff
@@ -0,0 +1,58 @@
+import { defaultSceneConfig, XRApp } from '@xrsdk/react'
+
+export function injectSceneHook() {
+  const sendMsg = (msg: any) => {
+    try {
+      //@ts-ignore
+      window.webkit.messageHandlers.bridge.postMessage(JSON.stringify(msg))
```
TrevorDev: this wont work on android, should use existing api for bridge instead

jsx-runtime/src/injectSceneHook.ts
mzvast: changed, and if #92 is merged it will works.

jsx-runtime/src/injectSceneHook.ts
mzvast: removed

PR #96, opened 2025-01-24
use addOnEngineUpdateEventListener instead of requestAnimationFrame t…

TrevorDev:
…o give webspatial more control over update loop

Currently apps can potentially miss frames if their update logic takes too long and they don't call requestAnimationFrame fast enuf, by changing the api, we can better handle situations like this in the future like driving the loop based on native code

### DIFF
--- a/runtime/src/core/SpatialSession.ts
+++ b/runtime/src/core/SpatialSession.ts
@@ -22,24 +22,21 @@ type animCallback = (time: DOMHighResTimeStamp) => void
  */
 export class SpatialSession {
   /** @hidden */
-  _animationFrameCallbacks = Array<animCallback>()
+  _engineUpdateListeners = Array<animCallback>()
   /** @hidden */
   _frameLoopStarted = false
 
   /**
-   * Request a callback to be called before the next render update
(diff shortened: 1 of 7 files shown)


runtime/src/core/SpatialSession.ts
```diff
@@ -22,24 +22,21 @@ type animCallback = (time: DOMHighResTimeStamp) => void
...
   /**
-   * Request a callback to be called before the next render update
-   * [TODO] might want to rework this to be triggered by native code or better handle frame timing
-   * @param callback callback to be called before next render update
+   * Add event listener callback to be called each frame
+   * @param callback callback to be called each update
    */
-  requestAnimationFrame(callback: animCallback) {
-    this._animationFrameCallbacks.push(callback)
+  addOnEngineUpdateEventListener(callback: animCallback) {
```
mzvast, line 33: Do we need a removeOnEngineUpdateEventListener, like addEventListener?

TrevorDev: Good point, can add it later

Issue #142, opened 2025-02-05
[Naming/Docs] Unclear what the behavior of these options are 'createRoot' | 'showRoot'


TrevorDev:
* @param {WindowStyle} [style='Plain'] - The style of the window to be created. Defaults to 'Plain'.
   * @param {Object} [cfg={}] - Configuration object for the window group. If provided, the window group will be standalone.
   * @param {Object} [cfg.sceneData] - Configuration for the scene data associated with the window group.
   * @param {string} [cfg.sceneData.method] - The method to be used for loading the scene.
   * @param {WindowGroupOptions} [cfg.sceneData.sceneConfig] - Configuration options for the scene.
   * @param {string} [cfg.sceneData.url] - The URL to load the scene from.
   * @param {string} [cfg.sceneData.windowID] - The ID of the window to be created.

Can more details be added to describe the behavior

Issue #112, opened 2025-01-26
some thoughts of scene


mzvast:
According to the discussion with @walton007 ,  there are some thoughts:

## scene manage

A scene can have states such as created (invisible), mounted (visible), and unmounted (destroyed), and the spatialWindowComponent might not be the most suitable for storing the state of the scene.

1. One approach is to attach an FSM class (e.g., sceneFSM) under each spatialWindowGroup to manage the scene logic.

2. Another approach is to introduce a sceneMgr to centrally manage the scenes.

## command

Currently scene-related commands has been moved under createWindowGroup command.

Scene-related commands should be separated out rather than being placed within createWindowGroup.

## atomic code or when it impossible to break down

Transfer as much native logic as possible to JavaScript, while retaining atomic capabilities in native code, to facilitate easier cross-platform compatibility. For example, move the logic for closing a scene to JS polyfill injection code.

However, if a function is only used by internal APIs and is sufficiently unique, making it impossible to break down further, then it might not be necessary to separate it out for use in JavaScript.

## your thinking

Put your comments, thanks:)

PR #161, opened 2025-02-06
chore: add SceneManager

mzvast:
Refactored some code, added the `createScene` command, and marked `session.createScene` as a hidden internal method.
Separated the scene logic in SpatialWindowComponent into `SceneManager`.

closing #142 #112

### DIFF
--- a/runtime/src/core/SpatialSession.ts
+++ b/runtime/src/core/SpatialSession.ts
@@ -188,7 +188,7 @@ export class SpatialSession {
    * @param {Object} [cfg={}] - Configuration object for the Scene.
    * @returns Boolean
    */
-  async createScene(
+  async _createScene(
     style: WindowStyle = 'Plain',
     cfg: {
       sceneData: sceneDataShape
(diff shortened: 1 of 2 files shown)


runtime/src/core/SpatialSession.ts
```diff
@@ -168,36 +173,28 @@ export class SpatialSession {
...
-   * @param {string} [cfg.sceneData.windowID] - The ID of the window to be created.
-   * @returns WindowGroup
+   * Creates a Scene to display content within an anchored area managed by the OS
+   * @hidden
+   * @param {WindowStyle} [style='Plain'] - The style of the Scene container to be created with. Defaults to 'Plain'.
+   * @param {Object} [cfg={}] - Configuration object for the Scene.
+   * @returns Boolean
    */
-  async createWindowGroup(
+  async createScene(
```
TrevorDev: Since this should be used via window.open(), can we make this @hidden and rename to _createScene?

mzvast: Good to follow convension😊

PR #163, opened 2025-02-06
Qa/songjing

493570584:


react/src/XRApp.ts
```diff
@@ -3,8 +3,8 @@ import { getSession } from './utils/getSession'
 
 export const defaultSceneConfig: WindowGroupOptions = {
   defaultSize: {
-    width: 900,
-    height: 700,
+    width: 100,
```
TrevorDev, line 6: Did you mean to change this?

PR #186, opened 2025-02-12
change  spatialdiv/html material's definition according to new requirement

walton007:
1. change material values to  none/transparent/translucent/thick/thin/regular
2. when html's material is set to none, it display as running in normal web browser
3. when spatialdiv's material is set to none, it behaves like transparent

### DIFF
--- a/builder/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift
+++ b/builder/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift
@@ -160,8 +160,6 @@ class SpatialWindowComponent: SpatialComponent {
                 // it's spatial div
                 webViewNative?.webViewHolder.appleWebView?.isOpaque = false
             }
-
-            print("webViewNative?.webViewHolder.appleWebView?.isOpaque", webViewNative?.webViewHolder.appleWebView?.isOpaque)
         }
     }
 


builder/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift
```diff
@@ -108,6 +109,10 @@ class SpatialWindowComponent: SpatialComponent {
         removeChildSpatialObject(spatialObject)
     }
 
+    private func isRootWebview() -> Bool {
```
TrevorDev, line 119: Can you document what is meant by root?

builder/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift
```diff
@@ -134,7 +139,24 @@ class SpatialWindowComponent: SpatialComponent {
...
-    var backgroundMaterial = BackgroundMaterial.None
+
+    private var _backgroundMaterial = BackgroundMaterial.None
+    var backgroundMaterial: BackgroundMaterial {
+        get {
+            return _backgroundMaterial
+        }
+        set(newValue) {
+            _backgroundMaterial = newValue
+            if isRootWebview() {
```
TrevorDev, line 157: Why do we need special logic here for root webview, can you document why this is needed? If we want spatial div to have different defaults, cant you do that in the JS code?

runtime/src/core/component/SpatialWindowComponent.ts
```diff
@@ -4,10 +4,11 @@ import { Vec3 } from '../SpatialTransform'
 
 export type BackgroundMaterialType =
```
TrevorDev, line 19: Can you document expected behavior of these in a comment?

builder/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift
walton007: we have comment for parentWebviewID in line 74
Line 74: // ID of the webview that created this or empty if its root
var parentWebviewID: String = ""

builder/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift
walton007: SpatialDiv and html have different behavior when backgroundMaterial  is None.
Currently we don't expose appleWebView.isOpaque property to js. So it's more easy to add the logic here.

runtime/src/core/component/SpatialWindowComponent.ts
walton007: document added

builder/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift
walton007: document added

mzvast: we already had `var isRoot = false` in Line 60. 
What is the difference to that root? Can we reuse that one?

walton007: is 'isRoot' the same as 'parentWebviewID =='' ' ?

mzvast: `isRoot` means it  is working like a browser tab or we called it scene.

walton007: have aligned with  @mzvast  that we'll keep private isRootWebview method and delete isRoot property

PR #215, opened 2025-02-17
support Model3D loading process event 

walton007:
1. add EventSpatialComponent to abstract the event handle logic
2. SpatialModel3DComponent support loading process like onSuccess/onFailure
3. SpatialInputComponent inherited from EventSpatialComponent


runtime/src/core/private/WebSpatial.ts
```diff
@@ -25,9 +27,27 @@ export class WebSpatial {
...
+    delete this.eventReceivers[resourceId]
+  }
+
   static init() {
     ;(window as any).__SpatialWebEvent = (e: any) => {
-      if (e.inputComponentID) {
+      if (e.resourceId) {
+        var callback = WebSpatial.eventReceivers[e.resourceId]
+        callback(e.data)
+      } else if (e.inputComponentID) {
```
TrevorDev, line 50: inputComponent here already has callback functionality. Can this be implemented in the same way as that instead of creating new classes/files?

testServer/src/model3d/index.tsx
```diff
@@ -85,11 +85,20 @@ function App() {
       <CSSModel3D
         ref={ref}
         style={styleOuter}
-        modelUrl="/src/assets/FlightHelmet.usdz"
+        modelUrl="/src/assets/FlightHelmets.usdz"
```
TrevorDev: why add the s here? If its to trigger error, can you call it fileThatDoesntExist.usdz?

builder/visionOSApp/web-spatial/libs/SpatialModel3DComponent.swift
```diff
@@ -13,6 +13,10 @@ class SpatialModel3DComponent: SpatialComponent {
     var resizable: Bool = true
     var aspectRatio: Double? = nil
 
+    // SpatialModel3DView should not have dependency on SpatialWindowComponent.
+    // It just need some JSB communication channel. This need to be refactor in future
```
TrevorDev, line 17: how would the refactor be done?

walton007: There'll be a independent JSBAgent with  message interface, any SpatialComponent and SpatialEntity can access the JSBAgent to send message to related webview runtime without aware of SpatialWindowComponent

testServer/src/model3d/index.tsx
walton007: yes. this is testing for trigger error. I can change the filename

runtime/src/core/private/WebSpatial.ts
walton007: I think WebSpatial.ts should not have any dependency on any specific Component. And SpatialInputComponent should  inhert from EventSpatialComponent (I would like to change SpatialInputComponent in another PR)

walton007: > inputComponent here already has callback functionality. Can this be implemented in the same way as that instead of creating new classes/files?

Do you want all interaction related logic wrapped in InputComponent?

TrevorDev: Ok, can you do the migration of inputComponent as part of this work aswell? I don't really like having 2 different paths for the same goal at the same time. If you want to break a pattern, suggest you update the pattern in the codebase first, then add your new feature with the new pattern.

builder/visionOSApp/web-spatial/libs/SpatialModel3DComponent.swift
TrevorDev: ok, can ignore this for now.

runtime/src/core/private/WebSpatial.ts
walton007: > Ok, can you do the migration of inputComponent as part of this work aswell? I don't really like having 2 different paths for the same goal at the same time. If you want to break a pattern, suggest you update the pattern in the codebase first, then add your new feature with the new pattern.

Migration of inputComponent have been finished in this PR

builder/visionOSApp/web-spatial/views/SpatialModel3DView.swift
```diff
@@ -32,18 +32,23 @@ struct SpatialModel3DView: View {
                     Model3D(url: url) { newPhase in
                         switch newPhase {
                         case .empty:
-                            ProgressView()
+                            EmptyView()
```
TrevorDev, line 35: Did you mean to change this?

walton007: At first, dexter wanted that developer could customize waitingUI instead of using progressView. 
but after some discussion, we decided still using ProgressView, So it'll be changed back to ProgressView in next PR.

PR #228, opened 2025-02-18
update SceneTest and depthPositionTest

493570584:


testServer/src/assets/Real-time_Bones_Demo_Phoenix_Bird.usdz
TrevorDev, line 1: Im not sure we should be checking in large files here as it bloats the library. Maybe we should host our assets on bytedance servers or using a public github pages? What do yall think @mzvast / @walton007 / @JackHFK

PR #235, opened 2025-02-19
reduce flash frequency when resizing windowgroup

mzvast:


builder/visionOSApp/web-spatial/views/PlainWindowContainerView.swift
```diff
@@ -71,7 +71,7 @@ struct PlainWindowContainerView: View {
                         timer!.invalidate()
                     }
                     // If we don't detect resolution change after x seconds we treat the resize as complete
-                    timer = Timer.scheduledTimer(withTimeInterval: 0.05, repeats: false) { _ in
+                    timer = Timer.scheduledTimer(withTimeInterval: 0.5, repeats: false) { _ in
```
TrevorDev: 10x longer o.O , could it be 0.2 maybe instead?

mzvast: Cool, 0.2 works good.

PR #237, opened 2025-02-19
chore: fix typo of readwindowgroupid

mzvast:


builder/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift
```diff
@@ -291,7 +291,7 @@ class SpatialWindowComponent: SpatialComponent {
         return targetUrl
     }
 
-    func readWinodwGroupID(id: String) -> String {
+    func readWindowGroupID(id: String) -> String {
```
TrevorDev: Name should switch to be readWindowContainerID?

mzvast: Changed, that make sense.

PR #238, opened 2025-02-19
fix material not render correct when back

mzvast:
## steps

click example

click back

see ghost render effect. No spatial material effect.

## before
image

## after
image

### DIFF
--- a/builder/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift
+++ b/builder/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift
@@ -358,9 +358,10 @@ class SpatialWindowComponent: SpatialComponent {
         isLoading = true
         loadingStyles = LoadingStyles()
 
-        // with materialWithBorderCorner
-        // SwiftUI not updating UIViewRepresentable when its deps update
-        // we trigger it by manually
+        // FIXME:
+        // This is a workaround to force run UIViewRepresentable.update()
+        // SwiftUI not trigger it when go back from example page.
+        // Warning of `AttributeGraph: cycle detected through attribute` fired when goes to example page
         webViewNative?.initialLoad()
     }
 


builder/visionOSApp/web-spatial/views/MaterialWithBorderCornerModifier.swift
```diff
@@ -70,6 +70,7 @@ struct MaterialWithBorderCornerModifier: ViewModifier {
             content
                 .clipShape(.rect(cornerRadii: radii))
         }
+        content
```
TrevorDev: Can you explain why this is the correct fix? Please add a comment as to why this is needed in the code too

mzvast: updated

TrevorDev: is there a radar or post online describing the issue? From the comment, im unsure how the bug works or if it will ever be resolved.

mzvast: No error prompt and only get come attribute graph cycle log, after that loading process is cut down, so JS code is not running through which leads to incomplete render.
The fix is to continue from where it left off.

mzvast: some other devs issue may help  https://stackoverflow.com/questions/62869586/how-do-i-debug-swiftui-attributegraph-cycle-warnings/63018486

mzvast: I found a similar bug.

https://developer.apple.com/forums/thread/124671

https://github.com/mediweb/UIViewRepresentableBug

mzvast: resolve for this workaround and track in #265

PR #241, opened 2025-02-19
fix dom sync on webview container resize

mzvast:
when open a new scene, document.body.clientWidth is zero and then it became some number greater than zero.

However this cannot be dectected by ResizeObserver, which leads to incorrect spatialDiv Translate.

## before
image

## after

image

### DIFF
--- a/react/src/spatial-react-components/SpatialReactComponent/StandardInstance.tsx
+++ b/react/src/spatial-react-components/SpatialReactComponent/StandardInstance.tsx
@@ -24,7 +24,9 @@ function useDetectDomRectChange() {
   })
 
   // listen to webview container size change and notifyDomChange
-  // this cannot be detected by ResizeObserver
+  // document.body.clientWidth may change as page loads.
+  // when creating a new scene, clientWidth changes from 0 to some positive value.
+  // this cannot be detected by ResizeObserver which only observe the dom element itself
   useEffect(() => {
     if (!ref.current || !spatialReactContextObject) {
       console.warn(


react/src/spatial-react-components/SpatialReactComponent/StandardInstance.tsx
```diff
@@ -23,6 +23,27 @@ function useDetectDomRectChange() {
     ref.current && spatialReactContextObject?.notifyDomChange(ref.current)
   })
 
+  // listen to webview container size change and notifyDomChange
+  // this cannot be detected by ResizeObserver
```
TrevorDev: Can you add a comment to why we need to know when the container size changed? Why does that impact the spatialdiv in a way that resizeObserver wouldnt catch it?

mzvast: comment added

TrevorDev: Oh i think it is related to both of these bugs https://github.com/spatial-web/XRSDK/issues/89 and https://github.com/spatial-web/XRSDK/issues/4 . I think the issue is that ResizeObserver doesn't detect position changes. Im not sure this fix actually solves the problem as it only catches translation bugs due to resize translation changes, not any translation changes.

TrevorDev: We might also need to use something like MutationObserver in XRSDK/testServer/src/docsWebsite/examples/spatialView.tsx but this may get fired very frequently so i would hesitate creating an observer for every SpatialComponent

mzvast: resize is not frequent event

PR #253, opened 2025-02-20
Create model react component

walton007:
model component is used to parse modelUrl and placeHolder.
use name "ModelNew" temporary, and will change to "Model" later

This PR is abased another PR https://github.com/spatial-web/XRSDK/pull/252

### DIFF
--- a/runtime/src/core/component/SpatialModel3DComponent.ts
+++ b/runtime/src/core/component/SpatialModel3DComponent.ts
@@ -20,7 +20,6 @@ export type TapEvent = {
  */
 export class SpatialModel3DComponent extends EventSpatialComponent {
   protected override onRecvEvent(data: any): void {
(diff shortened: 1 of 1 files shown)


runtime/src/core/component/SpatialModel3DComponent.ts
```diff
@@ -20,7 +20,7 @@ export type TapEvent = {
  */
 export class SpatialModel3DComponent extends EventSpatialComponent {
   protected override onRecvEvent(data: any): void {
-    console.log('onRecvEvent', data)
+    // console.log('onRecvEvent', data)
```
TrevorDev: why not just delete it?

walton007: When I test interaction related feature, I always need to print the data for debug

PR #268, opened 2025-02-25
JS packing: dynamic select sdk version based on process.env.XR_ENV, with esbuild/vite/webpack/rollup define plugin included

mzvast:
## What is this

We want devs can build different package targeting specific platform like web or avp.
We can ship 2 versions of react sdk in xrsdk-switch-xx branch already.
This PR changed the way we select subpackage.

## Why

Previously we select version by change the npm's package.json export field, which is universal and simple to maintain.

However, it was quite ugly because it changes the node_modules content.

To not modify node_modules, what we can do it to add one entry and in there use require to swich subpackage based on process.env.xxx.

I really don't like developers to write some plugin or tweak their build config so that they may not very happy to adopt our product. So I created some plugins which can set these define variables for them, covering most used packing tools today-- esbuild/vite/rollup/webpack. Devs can choose to use one into their build process.

### DIFF
--- a/react/esbuild.mjs
+++ b/react/esbuild.mjs
@@ -18,22 +18,6 @@ function clearDist(dir) {
 
 // for dynamic choose version we should ship 2 kinds of cjs output and switch by env variable
 
-// for developer to use, we should ship 2 kinds of output
-// 1. ESM unbundled, used with compiling tools (this not treeshaking well for unused import)
-// 2. ESM bundled, used by <script> tag
-
-// to target platforms we provide:
-// 1. default. AVP impl
-// 2. web.
-
-// so we have 2*2 matrix as follow:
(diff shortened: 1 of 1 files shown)


react/esbuild.mjs
```diff
@@ -70,6 +72,43 @@ const targets = [
     ],
     alias: { '@xrsdk/runtime': './src/noRuntime.ts' }, // replace the reference to runtime with empty module
   },
+  // {
```
TrevorDev: Remove commented out code?

react/esbuild.mjs
```diff
@@ -103,6 +142,9 @@ async function run() {
     contexts.push(ctx)
   }
 
+  // mkDistDir()
```
TrevorDev: remove

PR #272, opened 2025-02-27
support more apis in webgl model tag backend

TrevorDev:


react/src/spatial-react-components/Model/index.tsx
```diff
@@ -115,20 +116,68 @@ function ModelBase(inProps: ModelProps, ref: ModelElementRef) {
...
-      if (props.contentMode !== undefined && props.contentMode !== 'fit') {
-        console.warn(
-          "Model element contentMode != fit isn't supported on 2D screens",
-        )
-      }
       myModelViewer.current!.addEventListener('load', event => {
         if (props.onLoad) {
-          props.onLoad({ target: null as any })
+          props.onLoad({
+            target: { ready: true, currentSrc: gltfSourceURL } as any,
```
TrevorDev, line 122: @walton007 i think the signature of onLoad looks pretty weird, why is ready and currentSrc inside target here but in drag events, its separate?

react/src/spatial-react-components/Model/index.tsx
```diff
@@ -115,20 +116,68 @@ function ModelBase(inProps: ModelProps, ref: ModelElementRef) {
...
+        }
+      })
+
+      myModelViewer.current!.addEventListener('pointermove', event => {
+        if (props.onDrag) {
+          props.onDrag({
+            eventType: 'drag',
+            translation3D: { x: 0, y: 0, z: 0 },
+            startLocation3D: { x: 0, y: 0, z: 0 },
+            target: (ref as any).current as ModelElement,
```
TrevorDev, line 144: @walton007 how am i supposed to get a modelElement from here? It seems like you dont do it either in native implementation.

react/src/spatial-react-components/Model/index.tsx
```diff
@@ -115,20 +116,68 @@ function ModelBase(inProps: ModelProps, ref: ModelElementRef) {
...
+        console.warn(
+          "Model element aspectRatio != 1 isn't supported on 2D screens",
+        )
+      }
+    }, [props.contentMode, props.resizable, props.aspectRatio])
+
     return (
-      <div className={className}>
+      <div ref={ref} className={className} style={style}>
         <model-viewer
```
TrevorDev, line 181: @walton007 @dexteryy what should i do with the modelviewer default drag behavior? I think we want the rotation event handling right? Is it ok if its on by default for webgl backend but not supportive by the native api?

react/src/spatial-react-components/Model/index.tsx
walton007: target is the dom element. not just an js object

react/src/spatial-react-components/Model/index.tsx
walton007: dragEvent.target points to the modelElement (which is a div)

react/src/spatial-react-components/Model/index.tsx
```diff
@@ -115,20 +116,68 @@ function ModelBase(inProps: ModelProps, ref: ModelElementRef) {
...
     }, [])
 
+    useEffect(() => {
+      if (props.contentMode !== undefined && props.contentMode !== 'fit') {
+        console.warn(
+          "Model element contentMode != fit isn't supported on 2D screens",
+        )
+      }
+      if (props.resizable !== undefined && props.resizable !== false) {
+        console.warn(
```
walton007, line 168: why giving a warning here? 
Developers may write code <Model contentMode /> that can be used both in 2D web and AVP. So he need to give the contentMode props value

react/src/spatial-react-components/Model/index.tsx
walton007: I think modelviewer could have a viewMode prop, which define the default drag behavior. 
Devs can prevent default behavior when handling dragEvent like event.preventDefault()

PR #344, opened 2025-03-06
add @webspatial/core-sdk as peer for @webspatial/react-sdk deps

mzvast:
Is it right?

### DIFF
--- a/react/package.json
+++ b/react/package.json
@@ -51,7 +51,7 @@
   },
   "peerDependenciesMeta": {
     "@webspatial/core-sdk": {
-      "optional": true
+      "optional": false
     },
     "react": {
       "optional": true


react/package.json
```diff
@@ -40,16 +40,19 @@
     "@webspatial/react-sdk": "workspace:*",
     "@google/model-viewer": "^4.0.0",
     "@types/three": "^0.170.0",
-    "@webspatial/core-sdk": "workspace:*",
     "lodash.isequal": "^4.5.0",
     "loglevel": "^1.9.1",
     "three": "^0.170.0"
   },
   "peerDependencies": {
+    "@webspatial/core-sdk": "workspace:*",
```
TrevorDev, line 48: I think this should be the real package, not the workspace version. (but im not sure if this might screw up local development)

react/package.json
```diff
@@ -40,16 +40,19 @@
...
     "react": ">=18.0.0",
     "react-dom": ">=18.0.0"
   },
   "peerDependenciesMeta": {
+    "@webspatial/core-sdk": {
+      "optional": true
```
walton007: is value of optional should be false?

mzvast: Right, if it should be installed, we should remove this field or set false.

mzvast: change to false

PR #355, opened 2025-03-07
check DOMContentLoaded in more robust way

mzvast:
there is some timing issue reated to loaded check. Page may missed the event but still need to eval function

### DIFF
--- a/react/src/jsx/injectSceneHook.ts
+++ b/react/src/jsx/injectSceneHook.ts
@@ -19,7 +19,6 @@ export async function injectSceneHook() {
   }
 
   onContentLoaded(async () => {
-    console.log('DOMContentLoaded is ready.')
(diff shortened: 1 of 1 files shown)


react/src/jsx/injectSceneHook.ts
```diff
@@ -5,9 +5,21 @@ export async function injectSceneHook() {
...
+      document.readyState === 'complete'
+    ) {
+      callback()
+    } else {
+      document.addEventListener('DOMContentLoaded', callback)
+    }
+  }
+
+  onContentLoaded(async () => {
+    console.log('DOMContentLoaded is ready.')
```
walton007: remove console.log

mzvast: removed

PR #370, opened 2025-03-10
fix: Resizing webpage seems to cause issues with <Model> #369

### LINKED ISSUE #369: Resizing webpage seems to cause issues with <Model>
Image

walton007:
fix #369

### DIFF
--- a/.changeset/six-emus-decide.md
+++ b/.changeset/six-emus-decide.md
@@ -1,0 +1,5 @@
+---
+"@webspatial/react-sdk": patch
+---
+
+fix: Resizing webpage seems to cause issues with <Model> #369


react/src/spatial-react-components/Model3D/useDetectLayoutDomUpdated.ts
```diff
@@ -3,6 +3,17 @@ import { useRef, useEffect } from 'react'
 export function useDetectLayoutDomUpdated(onDomUpdated: () => void) {
   const ref = useRef<HTMLDivElement>(null)
 
+  // detect window resize
+  useEffect(() => {
+    const handleResize = () => {
+      onDomUpdated()
+    }
+    window.addEventListener('resize', handleResize)
```
mzvast, line 11: Will remove this listener cause not react to container size change?
I remember this was related to another bug.

TrevorDev: I am worried that resize is a subset of the changes we are not detecting. I thought we already added a resize handler somewhere else. Hope we dont add a bunch of one off event listeners being added everywhere doing similar things.

TrevorDev: This change looks OKish to me but if theres a better fix please do that instead. Will leave this to @mzvast to approve.

walton007: Maybe a resize hook can be used to reduce duplicate code

PR #381, opened 2025-03-10
when load failure occurs, we should fire an onLoad event with ready: …

TrevorDev:
…false

### DIFF
--- a/.changeset/calm-horses-peel.md
+++ b/.changeset/calm-horses-peel.md
@@ -1,0 +1,5 @@
+---
+"@webspatial/react-sdk": patch
+---
+
+when load failure occurs, we should fire an onLoad event with ready: …


react/src/spatial-react-components/Model/index.tsx
```diff
@@ -133,6 +133,17 @@ function ModelBase(inProps: ModelProps, ref: ModelElementRef) {
       if (!modelViewerExists) {
         return
       }
+
+      myModelViewer.current!.addEventListener('error', event => {
+        if ((event as any).detail.type == 'loadfailure') {
+          if (props.onLoad) {
+            props.onLoad({
+              target: { ready: false, currentSrc: gltfSourceURL } as any,
```
walton007, line 141: target should be a ModelElement which inherit from Div

PR #390, opened 2025-03-11
change ModelDragEvent to NativeModelDragEvent in coresdk

walton007:


react/src/spatial-react-components/Model3D/Model3DNative.ts
```diff
@@ -1,5 +1,5 @@
 import {
-  ModelDragEvent,
+  NativeModelDragEvent,
```
TrevorDev: @walton007 why does this name start with Native but the others start with Spatial. Maybe this should be SpatialModelDragEvent?

walton007: SpatialModelDragEvent seems a better name than NativeModelDragEvent. I can change it to NativeModelDragEvent

PR #402, opened 2025-03-13
[scene][part-04] EntityComponent support shift to newContainer

mzvast:


runtime/src/core/private/WebSpatial.ts
```diff
@@ -152,6 +153,22 @@ export class WebSpatial {
     }
   }
 
+  static async closeScene(container: SpatialWindowContainer) {
+    var cmd = new RemoteCommand('closeScene', {
+      windowContainerID: container._wg.id,
+    })
+
+    try {
```
TrevorDev: why does this need a try catch?

react/src/XRApp.ts
```diff
@@ -137,6 +140,35 @@ export class XRApp {
...
+          method: 'createRoot',
+          window, // pass self
+          // windowID: (newWindow as any)._webSpatialID,
+          // windowContainerID: (newWindow as any)._webSpatialGroupID,
+        },
+      },
+    )
+  }
+
+  async closeRootScene() {
```
TrevorDev, line 160: why do we need a specific command just for the root scene?

runtime/src/core/private/WebSpatial.ts
mzvast: removed

react/src/XRApp.ts
mzvast: we reused the _createScene cmd and the only difference is the rootScene's window has no `window._webSpatialID` value.

PR #403, opened 2025-03-13
[scene][part-05] support dynamic load first scene

mzvast:


cli/template/visionOSApp/web-spatial/libs/SpatialWindowComponent.swift
```diff
@@ -167,6 +167,8 @@ class SpatialWindowComponent: SpatialComponent {
 
     var didFailLoad = false
 
+    var isDynamicRoot = false // mark for dynamic loading the 1st scene
```
TrevorDev: this seems like a variable that will be hard to maintain, would be good to document this behavior more. LIke when is dynamic loadiing needed? Why isn't it always true, how does this work, etc.

mzvast: comments updated

PR #458, opened 2025-03-21
add entry for jsx-runtime to compatible with hard coded framework infer, eliminate the need for more alias

mzvast:


react/package.json
```diff
@@ -14,7 +14,11 @@
     "./plugin-rollup.js": "./npm/plugin/rollup.js",
     "./plugin-webpack.js": "./npm/plugin/webpack.js",
     "./jsx-runtime": "./dist/jsx/jsx-runtime.js",
-    "./jsx-dev-runtime": "./dist/jsx/jsx-dev-runtime.js"
+    "./jsx-dev-runtime": "./dist/jsx/jsx-dev-runtime.js",
+    "./default/jsx-runtime": "./dist/jsx/jsx-runtime.js",
+    "./default/jsx-dev-runtime": "./dist/jsx/jsx-dev-runtime.js",
+    "./web/jsx-runtime": "./dist/jsx/jsx-runtime.js",
```
TrevorDev, line 20: This web vs default stuff is getting pretty complicated 😵‍💫

mzvast: Absolutely agree, adding entries here it is not correct. I fixed it in plugin instead. :)

PR #537, opened 2025-03-27
update logic about outDir and base

mzvast:


react-plugin-vite/src/index.ts
```diff
@@ -26,9 +30,8 @@ export default function (options: WebSpatialOptions = {}): PluginOption[] {
             },
           },
         }
-
+        config.base = finalBase
         if (mode === AVP) {
-          config.base = '/webspatial/avp/'
```
walton007, line 31: i hope this path could be configiruable by devs

mzvast: It can be changed by user config

PR #549, opened 2025-03-31
Support xr order

walton007:


.changeset/four-buckets-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@webspatial/react-sdk': patch
+---
+
+support --xr-z-index
```
mzvast, line 5: Can you add this property to react/jsx/jsx-namespace.ts as well? So the tsc will not complain about missing it

walton007: it can be added in this MR https://github.com/webspatial/webspatial-sdk/pull/554

walton007: react/jsx/jsx-namespace.ts have been updated

PR #554, opened 2025-03-31
update react-sdk dts

mzvast:


react/src/jsx/jsx-namespace.ts
```diff
@@ -95,6 +95,15 @@ export namespace WebSpatialJSX {
...
+    [K in keyof ReactJSXIntrinsicElements]: ReactJSXIntrinsicElements[K] & {
+      style?: React.CSSProperties
+    }
+  }
+}
+
+declare module 'react' {
+  interface CSSProperties {
+    '--xr-background-material'?: string
+    '--xr-back'?: number | string
```
walton007, line 107: we have another custom css property named  "--xr-z-index".

mzvast: As the time of this PR commit, we don't have --xr-z-index yet. Would be clear to add in #549 .

Issue #613, opened 2025-04-11
Autocomplete not working as expected


TrevorDev:
When trying to use SpatialHelper, autocompletes to the packagename/dist folder, should just be packagename

PR #614, opened 2025-04-11
[Part 1] fix core-sdk import autocomplete to just be packagename instead of packagename/dist

TrevorDev:
https://github.com/webspatial/webspatial-sdk/issues/613

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -4,11 +4,10 @@
   "description": "this is the core js API for webspatial",
   "main": "dist/index.js",
   "module": "dist/index.js",
-  "types": "dist/index.d.js",
+  "types": "dist/index.d.ts",
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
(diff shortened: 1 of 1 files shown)


core/package.json
```diff
@@ -3,7 +3,15 @@
   "version": "0.1.10",
   "description": "this is the core js API for webspatial",
   "main": "dist/index.js",
+  "module": "dist/index.js",
   "types": "dist/index.d.js",
```
mzvast: types should end with d.ts as well

core/package.json
```diff
@@ -3,7 +3,15 @@
...
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "require": "./dist/index.js",
```
mzvast: require is for cjs format, but this file is ESM format.
do we need a cjs version? Or just remove it.

Issue #233, opened 2025-02-19
Not sure i agree of existence of isRootWebview or its usage


TrevorDev:
```
// Why are we changing background behavior based on this

  if isRootWebview() {
        webViewNative?.webViewHolder.appleWebView?.isOpaque = _backgroundMaterial == .None
    } else {
        // it's spatial div
        webViewNative?.webViewHolder.appleWebView?.isOpaque = false
    }



// Should this be enabled with a different prop instead (eg. enableNavView)

 if wv.isRootWebview() {
      NavView(swc: wv)
  }
```

PR #621, opened 2025-04-11
check if its root by if its entity is in root space

TrevorDev:
https://github.com/webspatial/webspatial-sdk/issues/233

### DIFF
--- a/builder/visionOS/web-spatial/libs/SpatialWindowComponent.swift
+++ b/builder/visionOS/web-spatial/libs/SpatialWindowComponent.swift
@@ -123,8 +123,8 @@ class SpatialWindowComponent: SpatialComponent {
     /// For example, a `SpatialDiv` is not a root webview.
     ///
     /// - Returns: `true` if the webview is a root webview (i.e., `parentWebviewID` is empty), otherwise `false`.
-    public func isRootWebview() -> Bool {
-        return parentWebviewID == ""
+    public func isRootEntityWebview() -> Bool {
+        return entity?.coordinateSpace == .ROOT
     }
 
     public func setWindowContainer(uuid: String, wgd: WindowContainerData) {
@@ -151,7 +151,7 @@ class SpatialWindowComponent: SpatialComponent {
         }
         set(newValue) {
             _backgroundMaterial = newValue
-            if isRootWebview() {
+            if isRootEntityWebview() {
                 webViewNative?.webViewHolder.appleWebView?.isOpaque = _backgroundMaterial == .None
             } else {
                 // it's spatial div
@@ -396,7 +396,7 @@ class SpatialWindowComponent: SpatialComponent {
 
     func didCloseWebView() {
         // if need
-        if isRootWebview() {
+        if isRootEntityWebview() {
             SceneManager.Instance.closeRoot(self)
         }
     }
--- a/builder/visionOS/web-spatial/views/SpatialWebViewUI.swift
+++ b/builder/visionOS/web-spatial/views/SpatialWebViewUI.swift
@@ -179,7 +179,7 @@ struct SpatialWebViewUI: View {
             .opacity(wv.opacity)
             .hidden(!ent.visible)
             .ornament(attachmentAnchor: .scene(.bottomTrailing), contentAlignment: .bottomTrailing) {
-                if wv.isRootWebview() && pwaManager.display != .fullscreen {
+                if wv.isRootEntityWebview() && pwaManager.display != .fullscreen {
                     NavView(swc: wv, navInfo: wv.navInfo)
                 }
             }


builder/visionOS/web-spatial/libs/SpatialWindowComponent.swift
```diff
@@ -151,7 +151,7 @@ class SpatialWindowComponent: SpatialComponent {
         }
         set(newValue) {
             _backgroundMaterial = newValue
-            if isRootWebview() {
+            if isRootEntityWebview() {
```
TrevorDev, line 154: @walton007 I still think the usage here is odd. Why is it changing material based on this? Can't the spatialDiv just set the material it wants manually in the js client code?

builder/visionOS/web-spatial/libs/SpatialWindowComponent.swift
```diff
@@ -123,8 +123,8 @@ class SpatialWindowComponent: SpatialComponent {
     /// For example, a `SpatialDiv` is not a root webview.
     ///
     /// - Returns: `true` if the webview is a root webview (i.e., `parentWebviewID` is empty), otherwise `false`.
-    public func isRootWebview() -> Bool {
-        return parentWebviewID == ""
+    public func isRootEntityWebview() -> Bool {
+        return entity?.coordinateSpace == .ROOT
```
mzvast, line 127: The difference between the 2 methods is
parentWebviewID is set when spawning ( createResource or create scene) and it will never change.

However, coordinateSpace can be modified by update resource, and change this espetially of a scene root swc will cause unexpected behaviors like not dismissing windowGroup on webview closed or have impact on show the navigation UI on it.

builder/visionOS/web-spatial/views/SpatialWebViewUI.swift
```diff
@@ -179,7 +179,7 @@ struct SpatialWebViewUI: View {
             .opacity(wv.opacity)
             .hidden(!ent.visible)
             .ornament(attachmentAnchor: .scene(.bottomTrailing), contentAlignment: .bottomTrailing) {
-                if wv.isRootWebview() && pwaManager.display != .fullscreen {
+                if wv.isRootEntityWebview() && pwaManager.display != .fullscreen {
```
walton007, line 182: I think the NavView has a strong connection with **Scene** concept. 
Only the Scene can have a NavView. 
Scene is not **EQUAL** to WindowGroup or (WindowGroup + Entity + WindowComponent).
Scene means to some  actions like **windowl.open**

If you create a WindowGroup by lowlevel api, the expected behavior is not defined

builder/visionOS/web-spatial/libs/SpatialWindowComponent.swift
walton007: do you want to expose  **"webViewNative?.webViewHolder.appleWebView?.isOpaque"** to js ? That would be more odd.
In fact, I prefer apparently define the Scene Concept (which have a root webview) and  SpatialDiv Concept in the native side apparently:
-  Scene defines a WindowContainer with a Webview Container to load the entry html
-  SpatialDiv define a Webview Container running JSX Div.

Although they both have backgroundMaterial, but with different meaning

TrevorDev: Why should webcontainers have different implementation on the native side here? We already provide a background material JS api. When you create a spatialDiv, you can default the background material value to whatever you think makes sense via JS.

builder/visionOS/web-spatial/views/SpatialWebViewUI.swift
TrevorDev: Scene is a high level concept right? 

In ThreeJS, it has higher level concepts like Mesh. Mesh is implemented on top of webGL which deals with buffers and shaders. It would be odd to add a webGL mesh object.

This is similar to our Scene. It can be implemented on top of (WindowGroup + Entity + WindowComponent). It would be odd to add a Scene concept to Native implementation.

If we keep going down this path we will end up with a lot of code to write on the native side for other platforms. Are we going to add a 3PanelScene next, or a carEntity native class? These are things that can be built using existing concepts.

walton007: 1. SceneManager logic is implemented in native. Is it possible to implement current window.open logic using pure js api on top of (WindowGroup + Entity + WindowComponent) ? @mzvast 
2. App lifecycle logic is related to a group of Scenes lifecycle from user's view. WindowGoup is part of a Scene. 
3. NavView is a native code and also part of a Scene. User can config a Scene to determine  the display of NavView.  
If user can create a WindowGroup, he will confused about how to config the NavView

builder/visionOS/web-spatial/libs/SpatialWindowComponent.swift
walton007: although both html material api and div have '--xr-background-material' css property, they have different behavior when have 'None' value.

PR #628, opened 2025-04-15
Fix spatialwindowcomponent setstyle

### LINKED ISSUE #282: Using image in spatialDiv causes page to go black
```
  <SpatialDiv spatialStyle={{
        position: { z: 50 },
        cornerRadius: 10,
      }}>
        test
        <img src=" https://picsum.photos/200/300" />
      </SpatialDiv>
```

Image

walton007:
fix #282

### DIFF
--- a/testServer/src/docsWebsite/examples/reactDomElement.tsx
+++ b/testServer/src/docsWebsite/examples/reactDomElement.tsx
@@ -1,19 +1,67 @@
 import { SpatialSession } from '@webspatial/core-sdk'
+import { useState } from 'react'
 import { showSample } from './sampleLoader'
 import { SpatialDiv } from '@webspatial/react-sdk'
+import { Euler, Quaternion } from 'three'
 
 function MySample(_props: { session?: SpatialSession }) {
+  var [rotation, setRotation] = useState(0)
   return (
-    <div className="flex flex-col items-center w-full">
-      <SpatialDiv
-        spatialStyle={{ position: { z: 50 } }}
-        style={{
-          maskImage:
-            'linear-gradient(to bottom, rgba(0,0,0,1.0), rgba(0,0,0,0))',
+    <div className="m-10">
+      <input
+        type="range"
+        step="0.005"
+        min={-Math.PI * 2}
+        max={Math.PI * 2}
+        className="mt-10 m-10 w-60 h-3 bg-gray-200 rounded-lg appearance-none cursor-pointer range-lg dark:bg-gray-700"
+        onChange={async e => {
+          setRotation(Number(e.target.value))
         }}
-      >
-        <img src="https://picsum.photos/200/300" />
-      </SpatialDiv>
+      ></input>
+      <div className="flex">
+        <SpatialDiv
+          spatialStyle={{
+            position: { z: 50 },
+            cornerRadius: 10,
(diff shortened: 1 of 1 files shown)


testServer/src/docsWebsite/examples/reactDomElement.tsx
```diff
@@ -1,67 +1,19 @@
...
-    <div className="m-10">
-      <input
-        type="range"
-        step="0.005"
-        min={-Math.PI * 2}
-        max={Math.PI * 2}
-        className="mt-10 m-10 w-60 h-3 bg-gray-200 rounded-lg appearance-none cursor-pointer range-lg dark:bg-gray-700"
-        onChange={async e => {
-          setRotation(Number(e.target.value))
+    <div className="flex flex-col items-center w-full">
```
TrevorDev: can you revert this file, this was just repro code overwriting this file.

builder/visionOS/web-spatial/libs/Utils/CommandManager.swift
```diff
@@ -492,9 +492,6 @@ class CommandManager {
             }
 
             if let backgroundMaterial: BackgroundMaterial = data.update?.style?.backgroundMaterial {
-                if target.isLoading {
```
TrevorDev, line 495: Can you summarize what the cause of this issue was? Some of this isLoading logic was here on purpose to avoid changing the current styles while the new webpage hasnt yet been displayed.

testServer/src/docsWebsite/examples/reactDomElement.tsx
walton007: OK. I'll revert the file

builder/visionOS/web-spatial/libs/Utils/CommandManager.swift
walton007: I have added some comment here: 
https://github.com/webspatial/webspatial-sdk/issues/282

walton007: **setStyle** is called by SpatialDiv and window global style.
They should have different behavior,  maybe we should provide different api

walton007: following process steps leads to the background material bug:
1. load entry html in root WindowComponent's webview
2. parse global css and call  setStyle(glassMaterial)
  - root WindowComponent's loadingStyles.backgroundMaterial is set to glassMaterial
3. parse spatialdiv
  - portalInstance will create a WindowComponent, let's called it SubWindowComponent
  -  portalInstance set SubWindowComponent's backgroundMaterial to None
  **- code in line 495 will loadingStyles.backgroundMaterial to None (This is not the expected behavior)**
 4. finish load entry html
   - didFinishLoadPage is called and loadingStyles.backgroundMaterial will be used to set the page's background material **which have wrong value 'None'**

TrevorDev: Discussed a bit:

basePage (MaterialType: None == css background #00000000 == white, opacity = true )
    spatialDiv (MaterialType: transparent == css background #00000000 == transparent, opacity=false)

basePage defaults to None
spatialDIv defaults to Transparent

PR #631, opened 2025-04-15
optimized will optimize standard instance memory usage, with one limi…

### LINKED ISSUE #156: Avoid duplicate elements somehow for heavy SpatialDiv content
For content like video or webgl, we should have some mechanism to avoid duplicating any elements.

walton007:
…tation that it's width/height need to be defined by css style

fix #156

### DIFF
--- a/react/src/jsx/jsx-namespace.ts
+++ b/react/src/jsx/jsx-namespace.ts
@@ -98,6 +98,7 @@ export namespace WebSpatialJSX {
     [K in keyof ReactJSXIntrinsicElements]: ReactJSXIntrinsicElements[K] & {
       style?: React.CSSProperties
       'enable-xr'?: boolean
+      optimized?: boolean
     }
   }
 }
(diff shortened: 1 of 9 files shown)


react/src/jsx/jsx-namespace.ts
```diff
@@ -98,6 +98,7 @@ export namespace WebSpatialJSX {
     [K in keyof ReactJSXIntrinsicElements]: ReactJSXIntrinsicElements[K] & {
       style?: React.CSSProperties
       'enable-xr'?: boolean
+      optimized?: boolean
```
TrevorDev, line 101: optimized isn't a great name, why would I want my div to not be optimized.

maybe one of these names could be better?

fixed-size, ignore-children-size, disable-duplicate-elements

walton007: 'ignore-children-size' is exactly what the spatialdiv optimization expect to know.
My only concern is that the name is a little long

walton007: @dexteryy need to know this PR. What's your opinion about this new 'disable-duplicate-elements' feature?

dexteryy: @walton007 I talked about it in the original issue.

PR #637, opened 2025-04-16
[part-01] export version from core-sdk

mzvast:


core/src/core/index.ts
```diff
@@ -8,3 +8,4 @@ export { SpatialHelper } from './SpatialHelper'
 export { type WindowStyle } from './types'
 export { SpatialTransform, Vec3, Vec4 } from './SpatialTransform'
 export * from './types'
+export const version = __coresdkversion__
```
walton007: is it possible to isolate version logic from the core logic?

PR #652, opened 2025-04-17
[part-02][native]get nativeAPIVersion from package version

mzvast:


builder/androidXRApp/webspatiallib/build.gradle.kts
```diff
@@ -7,12 +8,34 @@ plugins {
...
+    }
 
     defaultConfig {
         minSdk = 34
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         consumerProguardFiles("consumer-rules.pro")
+
+        // 1. read visionOS/package.json
+        val pkgFile = file("$rootDir/../visionOS/package.json")
```
TrevorDev: Maybe we should create a package.json for android? Or create a github issue to track fixing this so we don't forget?

mzvast: added a package.json for androidxrapp. It will bump version as it but not publish npm.

builder/visionOS/web-spatial/libs/Utils/version.swift
```diff
@@ -0,0 +1,26 @@
...
+        .deletingLastPathComponent() // .. project root
+        .appendingPathComponent("package.json")
+
+    do {
+        // read file
+        let data = try Data(contentsOf: packageJSONURL)
+        let package = try JSONDecoder().decode(PackageJSON.self, from: data)
+        return package.version
+    } catch {
+        print("⚠️ Error reading package.json: \(error)")
```
TrevorDev: why would this fail? Should we crash here instead of returning nil?

pnpm-workspace.yaml
```diff
@@ -5,7 +5,7 @@ packages:
   - 'cli'
   - 'examples/*'
   - 'plugins/*'
-  - 'builder/visionOS'
+  - 'builder/*'
```
TrevorDev, line 8: cc @JackHFK fyi

builder/visionOS/web-spatial/libs/Utils/version.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+struct PackageJSON: Codable {
+    let version: String
+}
+
+func getPackageVersion() -> String? {
+    let currentFileURL = URL(fileURLWithPath: #file) // path of version
```
TrevorDev, line 8: is this going to run every startup of the app? Will this cause a long delay?

builder/visionOS/web-spatial/libs/Utils/version.swift
mzvast: Good to fail fast:)

builder/visionOS/web-spatial/libs/Utils/version.swift
mzvast: I tested the time cost of parsing package.json. The average time cost on my machine is 2.7ms, it seems acceptable.

```
func timedGetPackageVersion() -> String {
    let start = DispatchTime.now()
    let version = getPackageVersion()
    let end = DispatchTime.now()
    
    let nanoDiff = end.uptimeNanoseconds - start.uptimeNanoseconds
    let ms = Double(nanoDiff) / 1_000_000
    print("⏱ getPackageVersion ：\(ms) ms")
    return version
}

let nativeAPIVersion = timedGetPackageVersion() //getPackageVersion()

(5.534917+
2.448375 +
1.60375 +
2.216125 +
2.006583)/5 = 2.76195 ms
```

PR #671, opened 2025-04-24
Support the use of the -- tryWithoutBuild=true parameter in the run

JackHFK:
… command to bypass compilation and run

- Now cli will compile using a copy of the xcode project instead of directly modifying the original project

### DIFF
--- a/cli/src/lib/cmds/build.ts
+++ b/cli/src/lib/cmds/build.ts
@@ -119,6 +119,14 @@ export async function run(args: ParsedArgs): Promise<boolean> {
   return true
 }
 
+/*
+  Every time the run command is executed,
+  the current command will be recorded through run_history.txt,
+  and when using the --tryWithoutBuild=true parameter,
+  it will be judged whether it is the same as the previous command. 
+  If it is the same, it will be defaulted as already compiled, 
+  and the compilation will be skipped and the application will be launched directly
+*/
 function checkRunHistory(cmd: string) {
   let historyFile = join(__dirname, '../../run_history.txt')
   if (!fs.existsSync(historyFile)) {


cli/src/lib/cmds/build.ts
```diff
@@ -88,6 +94,21 @@ export async function store(args: ParsedArgs): Promise<boolean> {
 // build and run on simulator
 export async function run(args: ParsedArgs): Promise<boolean> {
   let appInfo = { name: 'WebSpatialTest', id: '' }
+  let paramArr = []
```
TrevorDev, line 97: Would encourage you to do a followup PR in the future to move to using https://www.npmjs.com/package/commander for arg parsing and command setup so you don't need to maining logic like this

cli/src/lib/xcode/xcrun.ts
```diff
@@ -123,11 +123,11 @@ export default class Xcrun {
       let cmd = new XcrunCMD().simctl()
       cmd.boot(device.deviceId)
       execSync(cmd.line)
+      // wait 10s for simulator to boot
+      execSync('sleep 10')
```
TrevorDev, line 127: is there a better way than hardcoding a 10 second delay?

cli/src/lib/cmds/build.ts
```diff
@@ -97,3 +118,28 @@ export async function run(args: ParsedArgs): Promise<boolean> {
   await XcodeManager.runWithSimulator(appInfo)
   return true
 }
+
+function checkRunHistory(cmd: string) {
+  let historyFile = join(__dirname, '../../run_history.txt')
```
TrevorDev, line 131: Can you add comments to this function. This history logic seems maybe a bit tricky to ensure it always works so we should be very clear what the intention of this code is for.

cli/src/lib/cmds/build.ts
```diff
@@ -88,6 +94,21 @@ export async function store(args: ParsedArgs): Promise<boolean> {
...
+  for (let key in args) {
+    if (key === '_') {
+      paramArr.push(`${args[key]}`)
+    } else if (key !== 'tryWithoutBuild') {
+      paramArr.push(`--${key}=${args[key]}`)
+    }
+  }
+  let runCmd = paramArr.join(' ')
+  // If this command is a new command, go through the build process; otherwise, go through the launch process
+  if (!checkRunHistory(runCmd) && args['tryWithoutBuild'] === 'true') {
```
TrevorDev, line 107: Why do we want a tryWithoutBuild flag instead of a skipBuild flag? Should the developer always use tryWithoutBuild? If we use skipBuild, we can move the check history logic onto the developer so we don't have to manage this logic ourselves. What do you think?

cli/src/lib/xcode/xcrun.ts
JackHFK: Due to the fact that starting the emulator takes a relatively long time, and the command line execution time is almost instantaneous, there is a certain probability of failure if the command to start the app is executed directly after the command line is completed (because the emulator is not ready), but currently there is no good way to know when the emulator is ready.
I have repeatedly tested it several times, and usually the emulator startup time exceeds 10 seconds, possibly several tens of seconds. In my testing, the command to launch the app can be executed correctly after 10 seconds.

cli/src/lib/cmds/build.ts
JackHFK: - Firstly, I believe that developers can implement similar logic themselves, as we provide the launch command.
- Secondly, in order to improve usage efficiency and avoid frequent compilation (such as in CI), I added the tryWithoutBuild parameter, which has two layers of meaning.
  1. Not using this parameter will not affect the previous user experience, as developers may really want to compile every time they use run.
  2. When using this parameter, if there is no compilation record before, it will still be compiled to ensure that the app is installed in the emulator, which is also the reason for adding "try", which means trying to skip it.

Perhaps you think we can change it to trySkipBuild?

cli/src/lib/cmds/build.ts
JackHFK: Ok, I will try using it first.

cli/src/lib/cmds/build.ts
walton007: Is it possible to always skip build again when possible?

cli/src/lib/cmds/build.ts
JackHFK: I added comments in https://github.com/webspatial/webspatial-sdk/pull/671/commits/ea982ff8fe1c800370566ffd50d9a377feb052de

PR #680, opened 2025-04-25
Replace the original command parsing logic with commander.js

JackHFK:


cli/src/lib/cmds/build.ts
```diff
@@ -77,23 +79,61 @@ export async function store(args: ParsedArgs): Promise<boolean> {
...
     }
     appInfo.name = buildRes.json.name
   }
   await XcodeManager.upload(args, appInfo)
-  return true
 }
 
 // build and run on simulator
-export async function run(args: ParsedArgs): Promise<boolean> {
+export async function run(args: any) {
```
TrevorDev, line 90: can this be typed instead of any?

PR #694, opened 2025-05-06
[builder avp]reset scene window.name when reopen app

mzvast:
qa found an issue that the window.name will persist when reopen, which will cause unwanted bahavior.

### DIFF
--- a/builder/visionOS/web-spatial/libs/SpatialWindowComponent.swift
+++ b/builder/visionOS/web-spatial/libs/SpatialWindowComponent.swift
@@ -227,7 +227,14 @@ class SpatialWindowComponent: SpatialComponent {
         webViewNative!.initialLoad()
     }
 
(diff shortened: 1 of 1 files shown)


builder/visionOS/web-spatial/libs/SpatialWindowComponent.swift
```diff
@@ -227,6 +227,13 @@ class SpatialWindowComponent: SpatialComponent {
         webViewNative!.initialLoad()
     }
 
+    // remove webview name, to address name issues when reopen the app
```
TrevorDev: Can you add a bit more of a description here of why this is needed? What is the expectation when the app is reopened? Is the issue that once the webview has a name it never updates? Is the issue that the root page should always have a blank name?

mzvast: updated :)

PR #708, opened 2025-05-12
[core, avp]support fixed scene and resizing restriction when initScene

mzvast:
# demo

https://github.com/user-attachments/assets/b189cbb5-b4e6-4672-b019-89fabd5bdbb8

# why add this
to support the requirements of demo app

# what it is

```js
 initScene('sa', () => ({
    defaultSize: {
      width: 900,
      height: 900,
    },
    resizability: {
      minWidth: 700,
      minHeight: 700,
      maxWidth: 700,
      maxHeight: 700,
    },
  }))
  winARef.current = window.open(
    'http://localhost:5173/src/scene/xrapp.html',
    'sa',
  )
```

when call `window.open` with resizeRange like above, we wil set scene not resizable. No drag handler in windowGroup.

```js
 initScene('sa', () => ({
    defaultSize: {
      width: 900,
      height: 900,
    },
    resizability: {
      minWidth: 700,
      minHeight: 700,
      maxWidth: 900,
      maxHeight: 900,
    },
  }))
  winARef.current = window.open(
    'http://localhost:5173/src/scene/xrapp.html',
    'sa',
  )
```

when call `window.open` with resizability like above, we wil set scene resizable between 700-900. Has drag handler in windowGroup.

# what it is not
main scene not supported yet

### DIFF
--- a/packages/cli/src/lib/pwa/config.ts
+++ b/packages/cli/src/lib/pwa/config.ts
@@ -141,7 +141,7 @@ export function configMainScene(manifestJson: Record<string, any>) {
       width: 1280,
       height: 1280,
     },
-    resizability: 'automatic',
+    resizability: 'nil',
   }
   if (manifestJson.xr_main_scene) {
     if (typeof manifestJson.xr_main_scene === 'object') {
@@ -157,7 +157,7 @@ export function configMainScene(manifestJson: Record<string, any>) {
         manifestJson.xr_main_scene.resizability,
       )
         ? manifestJson.xr_main_scene.resizability
-        : 'automatic'
+        : 'nil'
 
       manifestJson.mainScene = mainScene
     } else if (typeof manifestJson.xr_main_scene === 'string') {


packages/visionOS/web-spatial/views/PlainWindowContainerView.swift
```diff
@@ -9,7 +9,21 @@ struct PlainWindowContainerView: View {
     @State private var timer: Timer?
 
     private func setSize(size: CGSize) {
-        sceneDelegate.window?.windowScene?.requestGeometryUpdate(.Vision(size: size))
+        sceneDelegate.window?.windowScene?
```
TrevorDev, line 12: Is this setSize api stable? I was seeing flakyness with it, in the past. Same goes with below, does it work 20/20 times in a row?

mzvast: As I tested by opening scene with initScene with fixed size,it works 20/20.

apps/test-server/src/scene/xrapp.tsx
```diff
@@ -134,6 +134,26 @@ function App() {
...
+      <button
+        className={btnCls}
+        onClick={async () => {
+          startlog('open')
+          initScene('sa', () => ({
+            defaultSize: {
+              width: 900,
+              height: 900,
+            },
+            resizability: 'contentSize',
```
TrevorDev: Should this be implemented internally using the _setOpenSettings api. I thought we agreed that scene stuff was going to be implemented on top of SpatialContainer api? Can we agree on https://bytedance.us.larkoffice.com/docx/RrDMdrgw8ocejtxzi3ruis6EsTf before merging this? 

cc @walton007

walton007: ok.
It's better we have a clear scene concept asap. @dexteryy

walton007 (approved):

apps/test-server/src/scene/xrapp.tsx
mzvast: I think we can merge this for now to unblock testing and refact in later PR

PR #722, opened 2025-05-14
[react-sdk] export dts for __XR_ENV_BASE__

mzvast:


packages/react/src/jsx/jsx-namespace.ts
```diff
@@ -117,4 +117,5 @@ declare global {
       defaultConfig: WindowContainerOptions,
     ) => Promise<WindowContainerOptions>
   }
+  var __XR_ENV_BASE__: string
```
walton007, line 120: I don't find any code about __XR_ENV_BASE__ in this codebase. Why add this declare here?

mzvast: because vite-plugin use it and  there is no place to declare it except here

PR #738, opened 2025-05-20
append git prefix to repository url

mzvast:
The repository.directory field in a package's package.json is particularly useful in monorepo projects. It clearly specifies the location of each subpackage within the repository, facilitating easier navigation and usage by tools and users alike. When using tools like pkg-pr-new, correctly configuring this field ensures accurate and effective preview deployments.

### DIFF
--- a/packages/androidXR/package.json
+++ b/packages/androidXR/package.json
@@ -8,7 +8,7 @@
   ],
   "repository": {
     "type": "git",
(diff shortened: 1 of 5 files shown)


TrevorDev:
@mzvast  Can you provide in the description why this is needed?

TrevorDev (approved):

PR #761, opened 2025-05-29
[ci]change tag name to vX.Y.Z after publishing

mzvast:


package.json
```diff
@@ -18,7 +18,7 @@
     "test": "concurrently \"pnpm -r --filter '@webspatial/*' run test\" \"tsc -p apps/test-server/tsconfig.json\"",
     "ciTest": "cd tests/ci-test/ && pnpm run test",
     "format": "pnpm -r run format",
-    "ci:publish": "changeset publish",
+    "ci:publish": "changeset publish --no-git-tag",
```
TrevorDev, line 21: why no tag?

.github/workflows/changesets.yml
```diff
@@ -52,22 +52,20 @@ jobs:
       - name: Create and push Git tags for published packages
         if: steps.changesets.outputs.published == 'true'
         run: |
-          echo "Published packages: ${{ steps.changesets.outputs.publishedPackages }}"
```
TrevorDev, line 55: why remove this?

TrevorDev (approved):

package.json
mzvast: by default it will generate tag like `package-name@v1.2.3`for every packages published ,but we hope to simplify by naming it `v1.2.3` ,because all our packages share the same version.

PR #804, opened 2025-06-27
config default manifest params with run command

JackHFK:


packages/cli/src/lib/pwa/index.ts
```diff
@@ -94,6 +96,15 @@ export class PWAGenerator {
     }
   }
 
+  private static compareManifest(manifest: Record<string, any>) {
```
mzvast, line 99: the func name compareXXX is confusing, accually it fill the defaults if not provide right?

PR #813, opened 2025-07-08
Fix resizability config

JackHFK:


mzvast (approved):

mzvast (commented):
QA found a problem that resizability in manifest file only work in the first run command, and changing the params and re-run will not take effect.  Maybe related to simulator.

PR #828, opened 2025-08-01
UpdateSceneConfigCommand

mzvast:


packages/core/src/refactor/JSBCommand.ts
```diff
@@ -126,19 +141,43 @@ export class PingCommand extends JSBCommand {
 
 /* WebSpatial Protocol Begin */
 abstract class WebSpatialProtocolCommand extends JSBCommand {
+  private windowProxyCallback?: (w: WindowProxy) => void
```
walton007: windowProxyCallback and setupWindowProxyCallback can be deleted.

PR #832, opened 2025-08-19
fix a tag hijack issue and add navigation listener

mzvast:
- onclick should return true when a tag not hijacked.

- capture scope event and trigger navigation, other stuff should go safari

### DIFF
--- a/packages/core/src/scene-polyfill.ts
+++ b/packages/core/src/scene-polyfill.ts
@@ -92,9 +92,12 @@ export function hijackWindowATag(openedWindow: WindowProxy) {
 
         // if onClick is set for the element, the raw onclick will be noop() trapped so the onclick check is no longer trustable
         // we handle all the scenarios
-        handleATag(e)
 
-        return false // prevent default action and stop event propagation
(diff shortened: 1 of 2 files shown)


packages/core/src/scene-polyfill.ts
```diff
@@ -115,6 +118,7 @@ function handleATag(event: MouseEvent) {
     if (target && target !== '_self') {
       event.preventDefault()
       window.open(url, target)
+      return true
     }
   }
```
walton007, line 123: why there is no return false?

packages/visionOS/web-spatial/model/SpatialScene.swift
```diff
@@ -85,6 +85,11 @@ class SpatialScene: SpatialObject, ScrollAbleSpatialElementContainer, WebMsgSend
         setupJSBListeners()
         setupWebViewStateListner()
     }
+    
+    private func handleNavigationCheck(_ url:URL) -> Bool {
+        // url in scope should open in place
```
walton007, line 90: @JackHFK  you can review this code

PR #844, opened 2025-09-08
Abstracting the SpatialObject protocol layer to make inheritance more…

JackHFK:
… flexible

### DIFF
--- a/packages/visionOS/web-spatial/SpatialObject.swift
+++ b/packages/visionOS/web-spatial/SpatialObject.swift
@@ -1,5 +1,14 @@
 import Foundation
 
+protocol SpatialObjectProtocol: Encodable, Equatable {
+    var spatialId: String { get }
+    var name: String { get set }
+    var isDestroyed: Bool { get }
+    
+    func destroy()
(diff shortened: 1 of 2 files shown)


packages/visionOS/web-spatial/SpatialObject.swift
```diff
@@ -8,23 +17,36 @@ class WeakReference<T: AnyObject> {
...
+    }
 
-    static func getWeakRef(_ id: String) -> SpatialObject? {
-        return weakRefObjects[id]?.value
+    static func getRefObject(_ id: String) -> SpatialObject? {
+        return SpatialObjectWeakRefManager.getWeakRef(id) as? SpatialObject
     }
 
-    let id: String
+    let spatialId: String
```
mzvast, line 47: Why another id? Is id ==  spatialId?

JackHFK: For better compatibility with inheritance of other classes, for example, Entity has its own ID and is not of type String, which directly leads to conflicts with the original ID and prevents inheritance

PR #862, opened 2025-09-28
chore: add webview fallback page

mzvast:


packages/visionOS/web-spatial/view/SpatialSceneContentView.swift
```diff
@@ -10,39 +10,53 @@ struct SpatialSceneContentView: View {
...
+                        Text("Failed to load webpage. Is the server running?")
+                            .foregroundColor(.white)
+                        Button("Reload") {
+                            let url = spatialScene.spatialWebViewModel.url
+
+                            spatialScene.spatialWebViewModel.load(url)
+                        }
+                        .foregroundColor(.white)
+                    }
+                    .frame(maxWidth: .infinity, maxHeight: .infinity).glassBackgroundEffect()
```
khufu-dev, line 24: Does this fill the parent container?

khufu-dev (approved):
Thanks for the cleanup work. Could we have a separate commit formatting all the code in Swift? That way code reviews will be focused on actual feature changes.

packages/visionOS/web-spatial/view/SpatialSceneContentView.swift
mzvast: yes

PR #867, opened 2025-10-15
fix reality event not fire

mzvast:


packages/core/src/SpatializedDynamic3DElement.ts
```diff
@@ -10,7 +10,11 @@ export class SpatializedDynamic3DElement extends SpatializedElement {
   rootEntity: SpatialEntity
   constructor(id: string) {
     super(id)
-    this.rootEntity = new SpatialEntity(id, { name: 'rootEntity' })
+    // the rootEntity is a fake entity to provide children only
+    // add suffix to avoid id conflict
+    this.rootEntity = new SpatialEntity(id + '_rootEntity', {
```
mzvast: rename to children

PR #869, opened 2025-10-17
Use local version of prettier when linting

khufu-dev:
All formatting commands use `npx prettier` except for the lint-staged command which runs as a pre commit hook. This requires prettier to be installed both locally and globally potentially causing a formatting mismatch due to version conflicts at commit time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,10 +29,10 @@
   },
   "lint-staged": {
     "*.{js,json}": [
-      "prettier --write"
+      "npx prettier --write"
     ],
     "*.ts?(x)": [
-      "prettier --write"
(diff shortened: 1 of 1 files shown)


mzvast:
I'm confused with `All formatting commands use npx prettier except for the lint-staged command` which I think isn't true .

we're using recommand config by [lint-staged.](https://github.com/lint-staged/lint-staged)

The key point is

1、Currently prettier is listed in devDependencies so it installed locally not globally.
Running lint-staged will use the version locally in node_modules.

2、prettier will use the .prettierrc to format so it should be of consistence across different versions.

PR #886, opened 2025-11-06
change android jsb to support promise

mzvast:
## android only
- add requestId in jsb message
- change callJSB to return promise and resolve after callback from native
- change JS interface  postMessage function signature to postMessage(requestId:string,cmd:string,data:string)

### DIFF
--- a/packages/core/src/platform-adapter/android/AndroidPlatform.ts
+++ b/packages/core/src/platform-adapter/android/AndroidPlatform.ts
@@ -34,8 +34,19 @@ export class AndroidPlatform implements PlatformAbility {
       try {
         const rId = nextRequestId()
 
+        SpatialWebEvent.addEventReceiver(rId, (result: JSBResponse) => {
+          SpatialWebEvent.removeEventReceiver(rId)
+          if (result.success) {
+            resolve(CommandResultSuccess(result.data))
+          } else {
+            const { code, message } = result.data as JSBError
+            reject(CommandResultFailure(code, message))
+          }
+        })
+
         const ans = window.webspatialBridge.postMessage(rId, cmd, msg)
         if (ans !== '') {
+          SpatialWebEvent.removeEventReceiver(rId)
           // sync call
           const result = JSON.parse(ans) as JSBResponse
           if (result.success) {
@@ -44,17 +55,6 @@ export class AndroidPlatform implements PlatformAbility {
             const { code, message } = result.data as JSBError
             resolve(CommandResultFailure(code, message))
           }
-        } else {
-          // async call
-          SpatialWebEvent.addEventReceiver(rId, (result: JSBResponse) => {
-            SpatialWebEvent.removeEventReceiver(rId)
-            if (result.success) {
-              resolve(CommandResultSuccess(result.data))
-            } else {
-              const { code, message } = result.data as JSBError
-              reject(CommandResultFailure(code, message))
-            }
-          })
         }
       } catch (error: unknown) {
         console.error(


packages/core/src/platform-adapter/android/AndroidPlatform.ts
```diff
@@ -4,24 +4,35 @@ import {
...
-      return CommandResultFailure(code, message)
-    }
+    return new Promise((resolve, reject) => {
+      try {
+        const rId = `rId${++requestId}`
+        // console.log(`${rId}::${cmd}::${msg}`)
+        SpatialWebEvent.addEventReceiver(rId, (data: any) => {
+          SpatialWebEvent.removeEventReceiver(rId)
+          resolve(CommandResultSuccess(JSON.parse(data)))
+        })
```
khufu-dev: In your commit comments can you explain why this is required on Android but not on visionOS. It's a significant architectural difference.

packages/core/src/platform-adapter/android/AndroidPlatform.ts
```diff
@@ -4,24 +4,35 @@ import {
...
+        window.webspatialBridge.postMessage(`${rId}::${cmd}::${msg}`)
```
khufu-dev: On Android there is support for directly calling Kotlin code. So we can define a `fun postMessage(requestID: Int, command: String, message: String): String` instead of merging everything into one string and parsing it on the Kotlin side.

packages/core/src/platform-adapter/android/AndroidPlatform.ts
```diff
@@ -4,24 +4,35 @@ import {
...
-      const result = await window.webspatialBridge.postMessage(`${cmd}::${msg}`)
-      return CommandResultSuccess(JSON.parse(result))
-    } catch (error: unknown) {
-      // console.error(`AndroidPlatform cmd: ${cmd}, msg: ${msg} error: ${error}`)
-      const { code, message } = JSON.parse((error as JSBError).message)
-      return CommandResultFailure(code, message)
-    }
+    return new Promise((resolve, reject) => {
+      try {
+        const rId = `rId${++requestId}`
```
khufu-dev: Since we're creating and then deleting the `requestId` after async function returns may I suggest using a  UUID instead of an incrementing counter. This ensures that we don't integer overflow when there are a large number of JSB calls.

packages/core/src/platform-adapter/android/AndroidPlatform.ts
```diff
@@ -6,30 +6,54 @@ import {
...
 }
 
 let creatingElementCount = 0
 
 let requestId = 0
 
+function addRequestId(msg: string, rId: string): string {
+  let updatedMsg = msg
+  if (msg && msg.trim() !== '{}') {
+    updatedMsg = msg.slice(0, -1) + `,"requestId":"${rId}"}`
```
khufu-dev: Can we avoid manipulating the string message? I believe it's much better to pass the requestId as a separate parameter and just handle it on the Kotlin `webspatialBridge.postMessage` without forwarding it to the rest of the system

khufu-dev (changes requested):

packages/core/src/platform-adapter/android/AndroidPlatform.ts
mzvast: comments added:)

packages/core/src/platform-adapter/android/AndroidPlatform.ts
mzvast: Cool, I created a 3 param fun

packages/core/src/platform-adapter/android/AndroidPlatform.ts
mzvast: In real world use case, if we emit 10000 query per seconds(a very crazy speed), it will take 9007199254740991/10000/3600/24/365 = `28,561` years to reach the max_safe_interger of javascript number.
 UUID is great solution but hard to see the sequence id and hard to debug.
So as a workaround I can just add a mod operator when creating the next id.

packages/core/src/platform-adapter/android/AndroidPlatform.ts
mzvast: Yeah, it is over complicated. 
removed

JackHFK (approved):

packages/core/src/platform-adapter/android/AndroidPlatform.ts
khufu-dev: I thought with 3D APIs and animations we could easily hit 10,000/s. But if that's a crazy speed your original solution of incrementing counter should be sufficient 😀

PR #887, opened 2025-11-06
Adding Geometry Primitives

parthpatelsj:
**Summary**
- Adds support for Plane, Sphere, Cone, and Cylinder geometries. Also improves geometry validation so JS/React receives clear errors instead of native crashes.

**Changes**
- Added SphereGeometry, ConeGeometry, CylinderGeometry, PlaneGeometry using MeshResource.generate*
- Updated Dynamic3DManager.createGeometry to construct all new types.
- SpatialScene.onCreateGeometry now validates required fields and returns typed errors (Box, Plane, Sphere, Cone, Cylinder).

**React**
- Added/updated React entities: PlaneEntity, ConeEntity, CylinderEntity, SphereEntity.
- Invalid geometry creation now surfaces clean error messages to JS.

**Test**
- Tested on visionOS
Screenshot 2025-12-02 at 2 20 07 PM

### DIFF
--- a/packages/core/src/Spatial.ts
+++ b/packages/core/src/Spatial.ts
@@ -13,10 +13,7 @@ export class Spatial {
    * [TODO] discuss implications of this not being async
    */
   requestSession() {
-    if (
-      this.isSupported() &&
-      this.getNativeVersion() === this.getClientVersion()
-    ) {
+    if (this.runInSpatialWeb()) {
       SpatialWebEvent.init()
       return new SpatialSession()
     } else {
@@ -36,17 +33,16 @@ export class Spatial {
     return false
   }
 
-  /**
+  /** @deprecated
    * Checks if WebSpatial is supported in the current environment.
    * Verifies compatibility between native and client versions.
    * @returns True if web spatial is supported by this webpage
    */
   isSupported() {
-    // todo: enhance judgement
-    return this.getNativeVersion() === this.getClientVersion()
+    return true
   }
 
-  /**
+  /** @deprecated
    * Gets the native WebSpatial version from the browser environment.
    * The version format follows semantic versioning (x.x.x).
    * @returns Native version string in format "x.x.x"
@@ -60,7 +56,7 @@ export class Spatial {
       : window.WebSpatailNativeVersion
   }
 
-  /**
+  /** @deprecated
    * Gets the client SDK version.
    * The version format follows semantic versioning (x.x.x).
    * @returns Client SDK version string in format "x.x.x"
--- a/packages/core/src/platform-adapter/index.ts
+++ b/packages/core/src/platform-adapter/index.ts
@@ -7,7 +7,10 @@ export function createPlatform(): PlatformAbility {
     return new SSRPlatform()
   }
 
-  if (window.navigator.userAgent.includes('Android')) {
+  if (
+    window.navigator.userAgent.includes('Android') ||
+    window.navigator.userAgent.includes('Linux')
+  ) {
     const AndroidPlatform = require('./android/AndroidPlatform').AndroidPlatform
     return new AndroidPlatform()
   } else {
--- a/packages/react/src/reality/components/BoxEntity.tsx
+++ b/packages/react/src/reality/components/BoxEntity.tsx
@@ -24,6 +24,7 @@ export const BoxEntity = forwardRef<EntityRefShape, BoxEntityProps>(
           height: props.height,
           depth: props.depth,
           cornerRadius: props.cornerRadius,
+          splitFaces: props.splitFaces,
         }}
       >
         {children}
--- a/packages/visionOS/web-spatial/manifest.swift
+++ b/packages/visionOS/web-spatial/manifest.swift
@@ -7,6 +7,7 @@ struct PWAManager: Codable {
     var isLocal: Bool = false
     var start_url: String = "http://localhost:5173/src/reality/geometryEntity"
 
+
     var scope: String = ""
     var id: String = "com.webspatial.pico"
 


apps/test-server/src/reality/geometry/index.tsx
```diff
@@ -0,0 +1,163 @@
+import React, { useRef, useState } from 'react'
+import ReactDOM from 'react-dom/client'
+import { getSession } from '@webspatial/react-sdk'
```
walton007: 'getSession' should not be available from react-sdk. This is a legacy bug. We'll fix it soon.
It seems that you have created testcase for core-sdk part, and missing react-sdk test code

packages/react/tsconfig.json
```diff
@@ -22,9 +22,15 @@
     "allowSyntheticDefaultImports": true,
     "jsx": "react-jsx",
     "paths": {
-      "@webspatial/react-sdk": ["./src"]
+      "@webspatial/react-sdk": ["./src"],
+      "@webspatial/core-sdk": ["../core/src"]
```
walton007: Is this required ?  Why adding this "@webspatial/core-sdk": ["../core/src"] ?

packages/react/src/types/globals.d.ts
```diff
@@ -0,0 +1,2 @@
+declare const __WEBSPATIAL_REACT_SDK_VERSION__: string
```
mzvast: It looks alreadly exist in `packages/react/src/types/global.d.ts`

packages/react/tsconfig.json
mzvast: I guess it may be for IDE code navigation? So as to jump directly into other package. Wonder if necessary

packages/react/src/types/globals.d.ts
mzvast: why do we need another `globals.d.ts` instead of using `global.d.ts`

packages/core/src/types/window.d.ts
```diff
@@ -0,0 +1,7 @@
+// Minimal ambient type so AndroidPlatform.ts compiles in non-Android envs
+declare global {
+  interface Window {
+    SpatialId?: unknown // keep optional & unknown to avoid leaking shape
+  }
+}
+export {}
```
khufu-dev: Why is this required? Maybe it should be a separate commit it's part of the Android work.

packages/react/src/reality/components/PlaneEntity.tsx
```diff
@@ -0,0 +1,88 @@
...
+      position,
+      rotation,
+      scale,
+      onSpatialTap,
+      children,
+      name,
+    },
+    ref,
+  ) => {
+    const ctx = useRealityContext()
```
khufu-dev, line 16: Best to make `ctx` not optional
```suggestion
    const ctx = useRealityContext()!
```

packages/react/src/reality/components/PlaneEntity.tsx
```diff
@@ -0,0 +1,88 @@
...
+              mesh: planeGeometry,
+              materials: materialList,
+            }),
+          )
+
+          await ent.addComponent(modelComponent)
+          return ent
+        } catch (error) {
+          await manager.dispose()
+          return null as any
```
khufu-dev: Why does this have to be cast as `any`?

packages/visionOS/web-spatial/manager/Dynamic3DManager.swift
```diff
@@ -1,50 +1,54 @@
...
+        guard let type = GeometryType(rawValue: props.type) else { return nil }
+        var geometry: Geometry? = nil
+        switch type {
         case .BoxGeometry:
-            BoxGeometry(width: props.width!, height: props.height!, depth: props.depth!, cornerRadius: props.cornerRadius ?? 0, splitFaces: props.splitFaces ?? false)
-        default:
-            nil
-        } else {
-            return nil
+            if let width = props.width, let height = props.height, let depth = props.depth {
```
khufu-dev: I think this changes the expected behaviour. Before we expected the width, height, and depth for a BoxGeometry and would otherwise crash but now you're silently ignoring them and returning `nil`.

khufu-dev (changes requested):

packages/react/tsconfig.json
parthpatelsj: Not necessary, can remove...

packages/react/src/types/globals.d.ts
parthpatelsj: I think I missed this, some reason my IDE was misconfigured.

packages/react/src/reality/components/PlaneEntity.tsx
parthpatelsj: I am following similar logic to BoxEntity, I believe it's to make tsc happy

packages/visionOS/web-spatial/manager/Dynamic3DManager.swift
parthpatelsj: My reasoning was that we should not fail in this layer, and instead have the JS/react layer fail, and return an error. However, if this was intended behavior I can revert this.

packages/core/src/types/window.d.ts
parthpatelsj: Hmm, when I try to push, it gives me an error if I don't have this: 
```adapter/android/AndroidPlatform.ts(57,25): error TS2339: Property 'SpatialId' does not exist on type 'Window'.```

(this is unrelated to this work though, so I would agree)

File: apps/test-server/src/reality/geometry/index.tsx (hunk 1 above)
Comment (parthpatelsj) [reply apps/test-server/src/reality/geometry/index.tsx]: This is now removed.

File: packages/visionOS/web-spatial/manager/Dynamic3DManager.swift (hunk 7 above)
Comment (parthpatelsj) [reply packages/visionOS/web-spatial/manager/Dynamic3DManager.swift]: If this logic is okay, I have gone ahead and added logic in SpatialScene to return failure (and clear message with missing fields). This will make the failure explicit and actionable in JS

File: packages/visionOS/web-spatial.xcodeproj/xcshareddata/xcschemes/web-spatial.xcscheme [hunk 8]
```diff
@@ -82,7 +82,7 @@
             isEnabled = "YES">
          </AdditionalOption>
          <AdditionalOption
-            key = "MallocScribble"
+            key = "MallocStackLogging"
```
khufu-dev: Do not commit.

packages/visionOS/web-spatial/manager/Dynamic3DManager.swift
khufu-dev: Ok fair point

packages/visionOS/web-spatial/model/SpatialScene.swift
```diff
@@ -39,7 +39,7 @@ class SpatialScene: SpatialObject, ScrollAbleSpatialElementContainer, WebMsgSend
     var parent: (any ScrollAbleSpatialElementContainer)?
 
     // Enum
-    public enum WindowStyle: String, Codable, CaseIterable {
```
khufu-dev, line 42: You don't want this accessible from outside this module?

packages/visionOS/web-spatial/model/SpatialScene.swift
```diff
@@ -778,12 +778,56 @@ class SpatialScene: SpatialObject, ScrollAbleSpatialElementContainer, WebMsgSend
...
+        if let gtype = GeometryType(rawValue: command.type) {
+            switch gtype {
+            case .BoxGeometry:
+                var missing: [String] = []
+                if command.width == nil { missing.append("width") }
+                if command.height == nil { missing.append("height") }
+                if command.depth == nil { missing.append("depth") }
+                if !missing.isEmpty {
+                    message = "missing required fields for BoxGeometry: " + missing.joined(separator: ", ")
+                }
```
khufu-dev: I don't understand this error handling. If these parameters are required shouldn't they be set on the React side (`SpatialBoxGeometryOptions`) as required? Why is everything set as optional on the React side but on the Swift side it throws an error?

Also it would be better to throw an Error in Dynamic3DManager.swift and simply forward the error message to JsbError instead of writing custom error messages for each Geometry type.

parthpatelsj: @mzvast - Is this something we can do (since you had written react side)? Then we can ensure TS gives right errors. 

I don't think it will hurt to have the check on native side, but can remove it if we don't think it's necessary. Have moved error throwing to `Dynamic3DManager.swift`

packages/visionOS/web-spatial/model/SpatialScene.swift
parthpatelsj: Good catch. My apologies, removing the public access level  was an oversight on my part during the refactor. That function needs to be public because it's part of the public API

packages/visionOS/web-spatial/manager/Dynamic3DManager.swift
```diff
@@ -1,50 +1,92 @@
...
-        default:
-            nil
-        } else {
-            return nil
+            var missing: [String] = []
+            if props.width == nil { missing.append("width") }
+            if props.height == nil { missing.append("height") }
+            if props.depth == nil { missing.append("depth") }
+            if !missing.isEmpty { throw GeometryCreationError.missingFields("BoxGeometry", missing) }
+            return BoxGeometry(width: props.width!, height: props.height!, depth: props.depth!, cornerRadius: props.cornerRadius ?? 0, splitFaces: props.splitFaces ?? false)
```
khufu-dev, line 40: I was thinking you can remove the above error messages and if one of the parameters were missing it would just throw an error on this line. We should be catching programming errors on the React side and shouldn't have too much error handling here.
But your current solution is good for now, thank you for making the changes.

packages/cli/src/lib/xcode/xcrun.ts
```diff
@@ -142,10 +142,9 @@ export default class Xcrun {
     console.log(`use simulator: ${device.deviceId}`)
     // launch visionOS simulator
     this.launchSimulator(device)
-    try{
+    try {
```
mzvast: Does this file only be formatted?  If so,  better to keep it as it was and ship formatting in other PR.

PR #889, opened 2025-11-11
Use 'uninstall' instead of 'terminate' to fix crash issues in specific situations

JackHFK:


packages/cli/src/lib/xcode/xcrun.ts
```diff
@@ -335,15 +339,23 @@ export default class Xcrun {
...
   private static parseDeviceInfo(device: string) {
-    const info = device.split('(')
-    const deviceInfo = {
-      name: info[0].trim(),
-      deviceId: info[1].split(')')[0].trim(),
-      state: info[2].split(')')[0].trim(),
-    }
-    return deviceInfo
+    const uuidRegex = /[A-F0-9]{8}-[A-F0-9]{4}-[A-F0-9]{4}-[A-F0-9]{4}-[A-F0-9]{12}/i;
+    const uuidMatch = device.match(uuidRegex) ?? ["()"];
```
khufu-dev, line 346: Why is the default value `"()"` instead of just `""`? Isn't the deviceId usually `"8C7AD003-4039-478F-9F94-938876D57817"`?

JackHFK: You're right, device IDs are usually in the form of uuids. After using the `xcrun simctl list devices 'Apple Vision Pro'`command, the usual format is `Apple Vision Pro (uuid) (Shutdown)`. What I can't be sure about is whether there will still be parentheses when uuid is empty, so I added a set of parentheses

PR #899, opened 2025-11-26
Add Sphere entity on visionOS

khufu-dev:
Created using Trae Solo using the following prompt:
> Look at PlaneEntity implementation and build out Sphere for visionOS.
> Look at the last 2 commits as a reference


apps/test-server/src/reality/geometry/index.tsx
walton007, line 1: getSession has been deprecated from 'react-sdk'. Try import from 'core-sdk'.

khufu-dev: Yes when it get's fixed in the [previous commit](https://github.com/webspatial/webspatial-sdk/pull/887#discussion_r2558687485) it will automatically be resolved here.

khufu-dev:
Cancelling since it's mainly merged into #887 and #902

PR #910, opened 2025-12-04
Add 3D model asset files

khufu-dev:
Add 3D model asset files. These are not too big and they are being used by our test files.


parthpatelsj (approved):

khufu-dev:
@walton007 merging 3D model files feel free to revert if this is a problem in the future.

PR #913, opened 2025-12-04
fix resource memory leak in 3d api

mzvast:


packages/react/src/reality/utils/ResourceRegistry.ts
```diff
@@ -11,11 +11,40 @@ export class ResourceRegistry {
...
+      // Schedule destruction when the resource becomes available
+      void p
+        .then(r => r.destroy())
+        .catch(e =>
+          console.error(
+            '[ResourceRegistry.removeAndDestroy] destroy failed:',
+            e,
+            id,
+          ),
+        )
```
khufu-dev: Can you please use modern `async await` Javascript?

packages/react/src/reality/utils/ResourceRegistry.ts
```diff
@@ -11,11 +11,40 @@ export class ResourceRegistry {
...
     this.resources.clear()
-    // todo: dothings to clear resources
+
+    // Best-effort destroy for all resolved and future-resolving resources
+    for (const p of pending) {
+      void p
+        .then(r => r.destroy())
+        .catch(e =>
+          console.error('[ResourceRegistry.destroy] destroy failed:', e),
+        )
```
khufu-dev:
```suggestion
    pending.forEach(promise => promise.then(spatialObj => spatialObj.destroy()))
```

Double check if the exception catching is necessary, I think the same thing will happen without you logging error.

packages/visionOS/web-spatial/model/dynamic3d/SpatialModelResource.swift
```diff
@@ -1,18 +1,18 @@
-import SwiftUI
 import RealityKit
+import SwiftUI
```
khufu-dev, line 2: Are these just formatting changes?

khufu-dev (approved):
How did you find this memory leak problem?

packages/visionOS/web-spatial/model/dynamic3d/SpatialModelResource.swift
mzvast: yes, only changed line24&30

mzvast:
> How did you find this memory leak problem?

1. enableDebugTool 
2. run inspectCurrentSpatialScene, see `spatialObjectRefCount`
3. refresh page
4. run inspectCurrentSpatialScene to compare results

```
// in js file
import {
  enableDebugTool,
} from '@webspatial/react-sdk'

enableDebugTool()
```

```
// in console
await inspectCurrentSpatialScene()
```
see logs
image

packages/react/src/reality/utils/ResourceRegistry.ts
mzvast: I've clean it a little bit.  
However, async signature is not sutable for this because in destroy stage of useEffect, react will not really await this to finish.

packages/react/src/reality/utils/ResourceRegistry.ts
mzvast: removed catch:)

PR #916, opened 2025-12-09
Entity Spatial Interaction

parthpatelsj:
Follow up to: https://github.com/webspatial/webspatial-sdk/pull/912, 

now dispatching entity-targeted drag/rotate/magnify gestures from SpatializedDynamic3DView.swift to the tapped/dragged entity’s spatialId

1. Spatial Drag (Translation) - $P_{new}$ = $P_{base}$ + $(T_{event}$ $\times$ k)
- $P_{new}$ is the new position $(x, y, z)$.
- $P_{base}$ is the position of the box when the drag *started* (stored in `dragBaseRef`).
- $T_{event}$ is the `e.detail.translation3D` vector from the event.
- $k$ is the `TRANSLATION_SCALE`
I've found k=0.001 is natural 

2. Spatial Rotate (Quaternion to Euler)
boxRot expects Euler angles, but rotation value encoded as quaternion

SwiftUI/RealityKit gives us the rotation as a quaternion (through JSON Encoding), but GeometryEntity only accepts Euler angles. So we take the quaternion from rotation.vector, convert it into roll/pitch/yaw, and then feed those Euler degrees back into the rotation field. It feels backwards — translation and scale stay in simple Vec3 form, but rotation comes in as a quaternion and goes out as Euler. It's basically mixing two rotation systems, which is why it feels inconsistent.

3. Spatial Magnify
$S_{new}$ = $S_{base}$ $\times$ $M_{event}$
- $S_{new}$ is the new scale vector $(x, y, z)$.
- $S_{base}$ is the scale when the gesture started.
- $M_{event}$ is the magnification factor.

Screenshot 2025-12-08 at 4 29 08 PM

### DIFF
--- a/apps/test-server/src/reality/gestures.tsx
+++ b/apps/test-server/src/reality/gestures.tsx
@@ -6,9 +6,6 @@ import {
   Entity,
   BoxEntity,
   UnlitMaterial,
-  SpatialTapEntityEvent,
-  SpatialDragEntityEvent,
-  SpatialRotateEntityEvent,
 } from '@webspatial/react-sdk'
 
 const btnCls =
@@ -17,6 +14,7 @@ const btnCls =
 function App() {
   const [logs, setLogs] = useState<string>('')
   const [enabled, setEnabled] = useState<boolean>(true)
+  const [exclusive, setExclusive] = useState<boolean>(false)
   const [matId, setMatId] = useState<'matGreen'>('matGreen')
   const [boxPos, setBoxPos] = useState({ x: 0, y: 0, z: 0 })
   const [boxRot, setBoxRot] = useState<{ x: number; y: number; z: number }>({
@@ -29,6 +27,7 @@ function App() {
   const dragBaseRef = useRef(boxPos)
   const scaleBaseRef = useRef(boxScale)
   const rotateBaseRef = useRef(boxRot)
+  const activeGestureRef = useRef<null | 'drag' | 'rotate' | 'magnify'>(null)
 
   function logLine(...args: any[]) {
     const msg = args
@@ -43,6 +42,9 @@ function App() {
       <div className="flex gap-2 my-3">
         <button className={btnCls} onClick={() => setEnabled(e => !e)}>
           {enabled ? 'Disable' : 'Enable'} Gestures
+        </button>
+        <button className={btnCls} onClick={() => setExclusive(e => !e)}>
+          {exclusive ? 'Exclusive' : 'Simultaneous'} Mode
         </button>
         <button className={btnCls} onClick={() => setLogs('')}>
           Clear Log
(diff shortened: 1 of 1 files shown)


packages/visionOS/web-spatial/view/SpatializedDynamic3DView.swift
```diff
@@ -7,90 +7,129 @@ struct SpatializedDynamic3DView: View {
...
+            if let entity = value.entity as? SpatialEntity {
+                if entity.enableRotate {
+                    let gestureEvent = WebSpatialRotateGuestureEvent(
+                        detail: .init(
+                            rotation: value.rotation,
+                            startAnchor3D: value.startAnchor3D,
+                            startLocation3D: value.startLocation3D
+                        )
+                    )
+                    spatialScene.sendWebMsg(entity.spatialId, gestureEvent)
```
mzvast, line 50: can we add onSpatialRotationStart here as well?

packages/visionOS/web-spatial/view/SpatializedDynamic3DView.swift
```diff
@@ -7,90 +7,129 @@ struct SpatializedDynamic3DView: View {
...
+                if entity.enableMagnify {
+                    let gestureEvent = WebSpatialMagnifyGuestureEvent(
+                        detail: .init(
+                            magnification: value.magnification,
+                            velocity: value.velocity,
+                            startLocation3D: value.startLocation3D,
+                            startAnchor3D: value.startAnchor3D
+                        )
+                    )
+                    spatialScene.sendWebMsg(entity.spatialId, gestureEvent)
```
mzvast, line 97: can we add onSpatialMagnifyStart here as well?

packages/visionOS/web-spatial/view/SpatializedDynamic3DView.swift
```diff
@@ -7,90 +7,129 @@ struct SpatializedDynamic3DView: View {
...
+                        detail: .init(
+                            location3D: value.location3D,
+                            startLocation3D: value.startLocation3D,
+                            translation3D: value.translation3D,
+                            predictedEndTranslation3D: value.predictedEndTranslation3D,
+                            predictedEndLocation3D: value.predictedEndLocation3D,
+                            velocity: value.velocity
+                        )
+                    )
+                    spatialScene.sendWebMsg(entity.spatialId, gestureEvent)
```
mzvast, line 148: can we add onSpatialDragStart here as well?

packages/visionOS/web-spatial/view/SpatializedDynamic3DView.swift
parthpatelsj: For this, I will need to add spatialdragstart, spatialrotatestart, and spatialmagnifystart to `WebMsgCommand.swift`

PR #921, opened 2025-12-11
fix window.open URL cast issue

mzvast:
what is the issue about:

If your are on `http://a.com`, then you click on a link to `http://b.com`, then it will add the prefix, and finally url become `http://a.comhttp://b.com`.

It should be `http://b.com`.


mzvast:
> Could you provide some more context on what was broken and how this is fixing it?

If your are on `http://a.com`, then you click on a link to `http://b.com`, then it will add the prefix, and finally url become `http://a.comhttp://b.com`.

It should be `http://b.com`.

PR #931, opened 2025-12-22
fix racing condition of modelAssets

mzvast:
# what is the problem

https://github.com/khufu-dev/webspatial-hackathon

There is a chance when opening the home page and do nothing and you'll see crash as bellow.

img_v3_02t7_76073b73-a662-41a4-a616-67073e73645g

and logs

```bash
NSDestinationFilePath=/Users/bytedance/Library/Developer/CoreSimulator/Devices/0F64EED5-1986-4260-A1DE-F47F9EECBBF8/data/Containers/Data/Application/62023F21-DB6E-4D8C-901B-A25CDCA9DF3E/Documents/vehicle-speedster.usdz, NSFilePath=/Users/bytedance/Library/Developer/CoreSimulator/Devices/0F64EED5-1986-4260-A1DE-F47F9EECBBF8/data/Containers/Data/Application/62023F21-DB6E-4D8C-901B-A25CDCA9DF3E/tmp/CFNetworkDownload_pmE4tb.tmp, NSURL=file:///Users/bytedance/Library/Developer/CoreSimulator/Devices/0F64EED5-1986-4260-A1DE-F47F9EECBBF8/data/Containers/Data/Application/62023F21-DB6E-4D8C-901B-A25CDCA9DF3E/tmp/CFNetworkDownload_pmE4tb.tmp, NSUnderlyingError=0x10264c300 {Error Domain=NSPOSIXErrorDomain Code=17 "File exists"}}
```

# reason

File move conflict: The “File exists” error in the logs indicates that when multiple threads download and attempt to move the same file simultaneously, a conflict occurs. One thread successfully moves the file, while the other threads fail because the destination file already exists. This failure triggers an error-handling path.

New data race: This error-handling path causes objects to be destroyed on a background thread, which in turn introduces another data race in SpatialObjectWeakRefManager, ultimately leading to the EXC_BAD_ACCESS crash you observed at the line weakRefObjects.removeValue(forKey: id).

# fixing

- adding a serial queue to guard the SpatialObject dictionary, ensuring that all accesses are synchronized.

- adding a serial queue to guard the weakRefObjects dictionary, ensuring that all accesses are synchronized.

- introduce a lock in Dynamic3DManager.swift to ensure that the sequence of “check if the file exists, delete it, and then move the file” is performed atomically, thereby preventing file conflicts.

### DIFF
--- a/packages/visionOS/web-spatial/manager/Dynamic3DManager.swift
+++ b/packages/visionOS/web-spatial/manager/Dynamic3DManager.swift
@@ -75,8 +75,8 @@ class Dynamic3DManager {
             loadComplete(.failure(NSError(domain: "Invalid URL", code: 0, userInfo: [NSLocalizedDescriptionKey: "Failed to create URL from string: \(urlString)"])))
             return
         }
-        var documentsUrl = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!
-        documentsUrl.appendPathComponent(url.lastPathComponent)
+        // Use an immutable URL to avoid capturing a mutable var in concurrent code
+        let documentsUrl = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!.appendingPathComponent(url.lastPathComponent)
         let session = URLSession(configuration: URLSessionConfiguration.default)
         var request = URLRequest(url: url)
         request.httpMethod = "GET"


packages/visionOS/web-spatial/manager/Dynamic3DManager.swift
```diff
@@ -97,6 +98,8 @@ class Dynamic3DManager {
             }
             let fileManager = FileManager.default
             do {
+                fileLock.lock()
```
khufu-dev: Can we do this without a lock, maybe use Swift concurrency? @parthpatelsj do you have any suggestions?

parthpatelsj: @khufu-dev makes a good point - if file operations takes long time, stalls the calling thread. Simplest way to use Swift concurrency in this case is with Actors. 
1) Create a file coordinator actor handles disk i/o (i.e. FileCoordinator)
2) Replace do/catch with `Task`

i.e. `try await FileCoordinator.shared.safelyMoveFile(from: location, to: documentsUrl)`

It will still be serial (only handle one file i/o at a time), but it's not blocking, only suspending (freeing thread to do other work)

Good article explaining
https://medium.com/@farshad.j/swift-actors-in-depth-19c8b3dbd85a

parthpatelsj:
I have added a single actor to serialize disk i/o, and refactored move in `Dynamic3DManager` to call FileCoordinator async. For creation/destroy entities, writes/removes are synchronized via serial queue.

PR #940, opened 2025-12-26
Add GLB models

khufu-dev:
Add GLB models for testing on Android devices. Enhance static 3D model test with error and load callback handlers.


parthpatelsj (approved):
Thanks for adding this

PR #946, opened 2026-01-14
fix: mute resource abort error log

mzvast:
In React strictMode, loading resource always will be interrupted to raise this error. It should be muted and nothing goes wrong

### DIFF
--- a/packages/react/src/reality/components/GeometryEntity.tsx
+++ b/packages/react/src/reality/components/GeometryEntity.tsx
@@ -84,7 +84,11 @@ export const GeometryEntity = forwardRef<EntityRefShape, GeometryEntityProps>(
           await ent.addComponent(modelComponent)
           return ent
         } catch (error) {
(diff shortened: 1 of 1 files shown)


packages/react/src/reality/components/GeometryEntity.tsx
```diff
@@ -84,7 +84,11 @@ export const GeometryEntity = forwardRef<EntityRefShape, GeometryEntityProps>(
           await ent.addComponent(modelComponent)
           return ent
         } catch (error) {
-          console.error(error)
+          if (error instanceof DOMException && error.name === 'AbortError') {
+            // AbortError is expected, just ignore it
```
khufu-dev, line 88:
```suggestion
            // In React strictMode, loading resource always will be interrupted to raise AbortError. It should be muted and nothing will go wrong
```

khufu-dev (approved):

PR #947, opened 2026-01-14
fix: mute bad model file log

mzvast:
bad model already handled in ModelAsset, no need to log again

### DIFF
--- a/packages/react/src/reality/components/ModelEntity.tsx
+++ b/packages/react/src/reality/components/ModelEntity.tsx
@@ -37,7 +37,8 @@ export const ModelEntity = forwardRef<EntityRefShape, Props>(
             { id, name },
           )
         } catch (error) {
-          console.error('ModelEntity error:', error)
+          // error already handled in ModelAsset, no need to log again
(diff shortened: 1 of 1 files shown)


packages/react/src/reality/components/ModelEntity.tsx
```diff
@@ -37,7 +37,8 @@ export const ModelEntity = forwardRef<EntityRefShape, Props>(
             { id, name },
           )
         } catch (error) {
```
khufu-dev, line 39: Can you remove the whole `try catch` and just return null when `modelAsset` is null?

khufu-dev (approved):

packages/react/src/reality/components/ModelEntity.tsx
mzvast: The `try catch` block is necessary because  `await ctx!.resourceRegistry.get(model)` may get a rejected promise so as to throw.
The `if(!modelAsset)`check is some sort of over protection, modelAsset should be a spatialObject or throw.

PR #950, opened 2026-01-21
update test-server

walton007:
- update home page information about spatialdiv and model usage
- update ReactSpringModel3DTest to use correct model asset
- add enableDebugTool for reality-test
- correct visibleTest case


khufu-dev (approved):
Please add screenshots of the changes made.

PR #952, opened 2026-01-22
fix unhandled error throw when destroy modelAssets

mzvast:
# the problem to fix

if the model url is not correct, createModelAsset cmd will rejected. 

When page unload, in destroy hook of useEffect, we cleaning up the pending promise which will raise the unhandled.

image

### DIFF
--- a/packages/react/src/reality/utils/ResourceRegistry.ts
+++ b/packages/react/src/reality/utils/ResourceRegistry.ts
@@ -18,7 +18,9 @@ export class ResourceRegistry {
     const p = this.resources.get(id)
     if (p) {
       // Schedule destruction when the resource becomes available
-      p.then(spatialObj => spatialObj.destroy())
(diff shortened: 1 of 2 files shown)


packages/react/src/reality/utils/ResourceRegistry.ts
```diff
@@ -18,7 +18,9 @@ export class ResourceRegistry {
     const p = this.resources.get(id)
     if (p) {
       // Schedule destruction when the resource becomes available
-      p.then(spatialObj => spatialObj.destroy())
+      p.then(spatialObj => spatialObj.destroy()).catch(() => {
```
khufu-dev, line 21: Do you want to ignore exceptions from both `p` and `spatialObj.destroy`?

mzvast: Yes, it will catch both.

PR #954, opened 2026-01-26
add unit test for core sdk and react sdk

walton007:


packages/core/src/coverage-boost.test.ts
```diff
@@ -0,0 +1,1050 @@
...
+      topLeading: 10,
+      topTrailing: 10,
+      bottomLeading: 0,
+      bottomTrailing: 20,
+    })
+  })
+
+  it('composeSRT composes translate and scale in expected order', () => {
+    if (!(globalThis as any).DOMMatrix) {
+      class DOMMatrixPolyfill {
```
khufu-dev, line 40: Is the DOMMatrix not available? Polyfill might result us testing against our own assumptions

packages/react/src/spatialized-container/hooks/useDomProxy.coverage.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  it('proxies style visibility/transform to task container', () => {
+    const ref = { current: null as any }
+    const proxy = new SpatialContainerRefProxy<any>(ref)
+    const dom = document.createElement('div')
+    const task = document.createElement('div')
+
+    proxy.updateStandardSpatializedContainerDom(dom)
+    proxy.updateTransformVisibilityTaskContainerDom(task)
+    const domProxy = ref.current as any
```
khufu-dev, line 29: Can we do without the `any` cast? It minimizes script Typescript.

PR #959, opened 2026-01-31
fix:  avoid .xr-spatial-default override user's css setting

walton007:
当使用webspatial SDK构建应用的时候，在PROD环境，CSS设置会不生效，定位原因是一些CSS设置被xr-spatial-default 覆盖了，所以这里修改下.xr-spatial-default的生效优先级，降为最低，避免覆盖用户设置的CSS样式

Also fixes css: `visibility`: 
- If hostDom.style.visibility is empty, set portal style.visibility to computed visibility
- Otherwise, don’t copy; portal stays visible by default

https://github.com/user-attachments/assets/b31a21c9-a457-475a-823e-727cfe52bd64


packages/react/src/spatialized-container/StandardSpatializedContainer.tsx
```diff
@@ -106,6 +106,8 @@ export function injectSpatialDefaultStyle() {
   const styleElement = document.createElement('style')
   styleElement.type = 'text/css'
   styleElement.innerHTML =
-    ' .xr-spatial-default {  --xr-back: 0; --xr-depth: 0; --xr-z-index: 0; --xr-background-material: none;  } '
+    ' :where(.xr-spatial-default) {  --xr-back: 0; --xr-depth: 0; --xr-z-index: 0; --xr-background-material: none;  } '
   document.head.appendChild(styleElement)
 }
+
+console.log(1111)
```
mzvast: remove if no use

PR #963, opened 2026-02-05
update 3d gesture param

mzvast:
both web and native side


packages/core/src/types/types.ts
```diff
@@ -357,9 +357,14 @@ export type SpatialDragEvent = CustomEvent<SpatialDragEventDetail>
 export type SpatialDragEndEvent = SpatialDragEvent
 
 export interface SpatialRotateEventDetail {
+  // TODO: should remove
   rotation: { vector: [number, number, number, number] }
```
khufu-dev, line 361: Why not remove it?

packages/core/src/types/types.ts
```diff
@@ -368,9 +373,6 @@ export type SpatialRotateEndEvent = SpatialRotateEvent
 
 export interface SpatialMagnifyEventDetail {
   magnification: number
-  velocity: number
-  startAnchor3D: Vec3
-  startLocation3D: Point3D
```
khufu-dev, line 373: Since this is an API change we need to have at least a minor or major version bump.

khufu-dev (changes requested):
Please update all tests that are affected by this change. Also add a changeset.

PR #981, opened 2026-02-18
Changeset base branch should be main

khufu-dev:
In daily development the base branch against which changes are compared should be the main branch. The stable branch is only used for releases to production.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -13,7 +13,7 @@
   ],
   "linked": [],
   "access": "public",
-  "baseBranch": "stable",
+  "baseBranch": "main",
   "updateInternalDependencies": "patch",
   "ignore": [],
   "privatePackages": { "version": true, "tag": false }


mzvast:
baseBranch

Because stable is the released branch, baseBranch should be set to stable to calculate changes against the published version.

Workflow

1. Develop feature branches from stable.

2. Merge features into main for integration.

3. Merge main into stable when releasing to npm.

khufu-dev:
> baseBranch
> 
> Because stable is the released branch, baseBranch should be set to stable to calculate changes against the published version.
> 
> Workflow
> 
> 1. Develop feature branches from stable.
> 2. Merge features into main for integration.
> 3. Merge main into stable when releasing to npm.

1. I believe most of our feature branches are based off of main branch since stable branch could be couple of weeks out of date. If you create a feature branch off of stable this is usually a bug fix.
2. Currently when you initiate changeset while working on a new feature off of main branch the tools marks all packages as changed since it's comparing against stable. This is incorrect and developer has to mark the correct packages as changed.

mzvast:
## To show this issue
 I run `npm run changeset` on branch main and I get some hints on which pkg change and if baseBranch is stable, it will mark all pkgs since last merging to stable as below screenshot.

### basebranch=stable
image

### basebranch=main
image

`changeset add` or short as `changeset` will read this `basebranch` from config and get changed pkg to inform us.

And changing basebranch will not have impact on `changeset version` command, which consume md files and bump pkg version.

## idea solution

we need get changed pkg hint from the branch we are merged to. 

That is to say, for most featues we are working we should target main. 

For hotfix on stable, we should target stable.

But this is dynamic setting basebranch is not supported.

I also noticed there was an issue mentioned same problem https://github.com/changesets/changesets/pull/1840.
They are using gitflow just like we do. The new verison cli added `--since` to override `basebranch` yesterday .

But this new version is not available for now. Maybe we can use this flag when it is ready.

## conclusion

I quite agree with you that we have a long period when main is ahead of stable. And setting basebranch to stable will have trouble creating changeset md files.
So it was good to optimize for daily changeset writting.

mzvast (approved):

Issue #926, opened 2025-12-18
Relative URLs don't work on spatialized Model tag


khufu-dev:
### Duplicate check

- [x] I searched existing issues and discussions

### Steps to reproduce

1. Place your model USDZ file in the `public/asset` folder
2. Paste the following code in your root webpage 
```tsx
<Model src="assets/Yamaha_TT-300_Record_Player.usdz" enable-xr />
```
3. Navigate to the root website [https://localhost:5173](https://localhost:5173)

## Expected
The model asset should load and display using the relative path. Similar to the non spatialized Model tag or using the root relative path `/assets/Yamaha_TT-300_Record_Player.usdz`. Also if there's an error loading the asset the `onError` callback should be triggered with a helpful error.

```tsx
<Model src="assets/Yamaha_TT-300_Record_Player.usdz" />
```

## Actual
The model does not load and no errors are given in the Javascript error.

### Dev Environment

• Run Target (Vision Pro / Simulator / Browser):  Simulator
• OS: macOS 26
• Node.js version: 24
• WebSpatial: latest
• Vite versions: 6

parthpatelsj:
Resolved with: https://github.com/webspatial/webspatial-sdk/pull/984

PR #984, opened 2026-02-23
fix: correct url resolution

parthpatelsj:
Fix for: https://github.com/webspatial/webspatial-sdk/issues/926

Switch to using native URL parser to automatically insert the missing forward slash between the base domain and relative assets/... path

Previously, we mashed the strings resulting in (i.e. https://localhost:5173assets/*.usdz...), whereas the new native URL parser understands routing and correctly builds (i.e. https://localhost:5173/assets/*.usdz) ... with the required slash.

### DIFF
--- a/.changeset/fix-model-url-resolution.md
+++ b/.changeset/fix-model-url-resolution.md
@@ -1,0 +1,5 @@
+---
+"@webspatial/react-sdk": patch
+---
+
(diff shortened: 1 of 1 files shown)


khufu-dev (approved):
Thank you, very elegant solution.

PR #988, opened 2026-02-24
sidebar late render triggers DOM update to spatial container

parthpatelsj:
- I added enable-xr-monitor to the app’s outer container div. 
- This scopes a single MutationObserver to the whole app subtree that includes both Sidebar and Main, so when the sidebar mounts and shifts layout, the monitor fires and the SDK recomputes the model’s placement.

https://webspatial.dev/docs/development-guide/using-the-webspatial-api/elevate-2d-elements#dynamic-change

Without this, sometimes, model loads out of place
Screenshot 2026-02-24 at 12 46 34 PM

### DIFF
--- a/apps/test-server/index.tsx
+++ b/apps/test-server/index.tsx
@@ -91,7 +91,7 @@ function App() {
         style={{ backgroundColor: 'var(--spa-bg-color, #ffffff)' }}
       >
         <Sidebar />
-        <main className={mainClass}>
+        <main className={mainClass} enable-xr-monitor>
           <Suspense
             fallback={
               <div className="flex items-center justify-center h-full">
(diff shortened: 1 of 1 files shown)


apps/test-server/index.tsx
```diff
@@ -89,6 +89,7 @@ function App() {
       <div
         className={outerClass}
         style={{ backgroundColor: 'var(--spa-bg-color, #ffffff)' }}
+        enable-xr-monitor
```
khufu-dev, line 92: Instead of doing this can you set a width on the sidebar? Not sure why the layout is jumping around as things load.

khufu-dev (changes requested):

apps/test-server/index.tsx
parthpatelsj: I can also set a width on sidebar, but the issue mentioned in the PR header will still remain without adding `enable-xr-monitor`.

khufu-dev: If it's a fixed sidebar then the layout shouldn't shift right?

khufu-dev:
Closing stale PR

PR #990, opened 2026-02-26
feat: AVP Support RealityView Attachments via <AttachmentAsset> + <AttachmentEntity>

parthpatelsj:
Implements “Attachments” feature for AVP (visionOS): render 2D HTML/React UI as 3D attachments that are parented under Reality entities and inherit their transforms.
Screenshot 2026-02-25 at 9 47 04 PM

### What’s included
- **React API**
  - `<AttachmentAsset name="...">` to declare attachment UI (outside `<SceneGraph>`)
  - `<AttachmentEntity attachment="..." position size />` to place UI in 3D under a parent entity
  - UI is rendered via `createPortal`, so attachment content shares React state with the main app
- **Core SDK**
  - Attachment types + `SpatialSession.createAttachmentEntity(...)`
  - Create/update/destroy commands (`webspatial://createAttachment`, `UpdateAttachmentEntity`, `DestroyCommand`)
  - `Attachment` wrapper with `getContainer()`, `update()`, `destroy()`
- **visionOS (AVP runtime)**
  - Native `AttachmentManager` + child `WKWebView` per attachment
  - `SpatialScene` intercepts `webspatial://createAttachment`, handles update/destroy, cleans up on reload/destroy
  - `SpatializedDynamic3DView` renders attachments via `RealityView(..., attachments:)` and parents them under the correct `SpatialEntity`
- **Navigation fix**
  - Prevents internal `webspatial://` URLs from being forwarded to `UIApplication.shared.open(...)`
- **Test page**
  - `/reality/attachments` demo: hide/show, animation (attachments follow parent), shared React state

### Notes
- Attachments are 2D UI surfaces only (no nested `<Reality>` / 3D APIs inside attachments).

### DIFF
--- a/docs/Attachments.md
+++ b/docs/Attachments.md
@@ -150,6 +150,6 @@ The recommended follow-up is to create a dedicated `AttachmentWebView: UIViewRep
 
 ## References
 
-- [RealityKit ViewAttachments](https://developer.apple.com/documentation/realitykit/primitives-in-realitykit#Add-attachments)
+- [RealityKit ViewAttachments](https://developer.apple.com/documentation/realitykit/attachment)
 - [React createPortal](https://react.dev/reference/react-dom/createPortal)
 - [WebSpatial Architecture — Intro to WebSpatial](https://webspatial.dev)


walton007:
特性功能方面的问题:
1. 是否支持多个AttachementEntity共享同一个AttachementAsset
2. 当前是否只支持inline style？ 是否需要像SpatialDiv一样支持页面的全局样式？
3. 当RealityView嵌套在SpatialDiv中的时候，AttachementEntity是否还能正常工作？
4. 当AttachementAsset中的jsx有Spatialdiv的时候，是否还能正常工作？
实现方面的问题:
1. AttachmentInfo 设计为继承自 SpatialObject是否更好？
2.  CoreSDK中是否应该直接暴露Attachment对象？是否因该遵循EC结构，比如只暴露Entity/AttachementComponent; 
  - CoreSDK 的 <Reality> 中是否能知有Entity/Component/Asset等核心概念的存在，这些核心概念映射到Native的核心概念中。 ReactSDK层面的<AttachmentEntity> 只是对底层EC Asset的快捷组装

apps/test-server/src/pages/reality/attachments.tsx
```diff
@@ -0,0 +1,289 @@
...
+    return () => {
+      if (animationRef.current) {
+        cancelAnimationFrame(animationRef.current)
+        animationRef.current = null
+      }
+    }
+  }, [animating])
+
+  const btnCls =
+    'select-none px-4 py-1 text-s font-semibold rounded-full border border-gray-700 hover:text-white bg-gray-700 hover:bg-gray-700 hover:border-transparent focus:outline-none focus:ring-2 focus:ring-gray-600 focus:ring-offset-2'
```
khufu-dev: Could this be simplified with https://daisyui.com

packages/react/src/utils/windowStyleSync.ts
```diff
@@ -0,0 +1,47 @@
...
+): Promise<boolean> {
+  return new Promise(resolve => {
+    n.href += '?uniqueURL=' + Math.random()
+    n.onerror = () => resolve(false)
+    n.onload = () => resolve(true)
+    setTimeout(() => {
+      childWindow.document.head.appendChild(n)
+    }, 50)
+  })
+}
```
khufu-dev, line 23: You seemed to have moved the code but deleted all the comments? It's helpful to have code refactor and formatting changes in a separate pull request so that's easier to review.

packages/react/src/spatialized-container/Spatialized2DElementContainer.tsx
```diff
@@ -224,6 +162,14 @@ function Spatialized2DElementContainerBase<P extends ElementType>(
   props: Spatialized2DElementContainerProps<P>,
   ref: ForwardedRef<SpatializedDivElementRef>,
 ) {
+  const insideAttachment = useInsideAttachment()
+  if (insideAttachment) {
+    console.warn(
+      '[WebSpatial] SpatialDiv cannot be used inside AttachmentAsset. Rendering as plain HTML.',
+    )
+    const { component: El, ...rest } = props as any
+    return <El ref={ref as any} {...rest} />
```
khufu-dev: Why is it being cast to `any`"? This could cause bugs later on.

packages/react/src/utils/windowStyleSync.ts
parthpatelsj: Good catch! Trae AI stripped the comments during the move without me noticing, but I've restored them.

walton007:
I prepared a minimal, API-compatible fix patch on branch `fix/attachments-style-sync` (commit `1b226c05`). 
                                                                                                               
    Key fixes (correctness/perf):                                                                              
    - Prevent unbounded <head> growth in attachment windows: tag synced nodes and clear previously-synced nodes
  before each re-sync.                                                                                         
      (packages/react/src/utils/windowStyleSync.ts)                                                            
    - Fix AttachmentEntity head-sync observer reliability: avoid using ref.current in effect deps, attach      
  observer based on a stable childWindow state, and debounce head sync by 100ms to reduce churn.               
      (packages/react/src/reality/components/AttachmentEntity.tsx)                                             
    - Handle runtime attachment name changes by migrating the registry mapping so portals don’t get stuck on   
  stale names.                                                                                                 
      (packages/react/src/reality/components/AttachmentEntity.tsx)                                             
    - Improve SpatialDiv degradation inside attachments: preserve original DOM props/component while stripping 
  spatial-only props to avoid React warnings.                                                                  
      (packages/react/src/spatialized-container/Spatialized2DElementContainer.tsx)                             
    - visionOS: if the parent entity is missing during update, re-parent the attachment back to the root entity
  (fallback consistent with initial make()).                                                                   
      (packages/visionOS/web-spatial/view/SpatializedDynamic3DView.swift)                                      
                                                                                                               
    This keeps the public API unchanged and specifically targets stability & performance risks discovered      
  during review.

parthpatelsj:
> I prepared a minimal, API-compatible fix patch on branch `fix/attachments-style-sync` (commit `1b226c05`).
> 
> ```
> Key fixes (correctness/perf):                                                                              
> - Prevent unbounded <head> growth in attachment windows: tag synced nodes and clear previously-synced nodes
> ```
> 
> before each re-sync. (packages/react/src/utils/windowStyleSync.ts) - Fix AttachmentEntity head-sync observer reliability: avoid using ref.current in effect deps, attach observer based on a stable childWindow state, and debounce head sync by 100ms to reduce churn. (packages/react/src/reality/components/AttachmentEntity.tsx) - Handle runtime attachment name changes by migrating the registry mapping so portals don’t get stuck on stale names. (packages/react/src/reality/components/AttachmentEntity.tsx) - Improve SpatialDiv degradation inside attachments: preserve original DOM props/component while stripping spatial-only props to avoid React warnings. (packages/react/src/spatialized-container/Spatialized2DElementContainer.tsx) - visionOS: if the parent entity is missing during update, re-parent the attachment back to the root entity (fallback consistent with initial make()). (packages/visionOS/web-spatial/view/SpatializedDynamic3DView.swift)
> 
> ```
> This keeps the public API unchanged and specifically targets stability & performance risks discovered      
> ```
> 
> during review.

Thanks for the fix, I have merged the branch into `feat/attachments`

apps/test-server/src/pages/reality/attachments.tsx
parthpatelsj: I have simplified the test

packages/react/src/utils/windowStyleSync.ts
```diff
@@ -0,0 +1,77 @@
...
+  openedWindow.document.body.style.minWidth = 'auto'
+  openedWindow.document.body.style.minHeight = 'auto'
+  openedWindow.document.body.style.maxWidth = 'fit-content'
+  openedWindow.document.body.style.minWidth = 'fit-content'
+  openedWindow.document.body.style.background = 'transparent'
+}
+
+export async function syncParentHeadToChild(childWindow: WindowProxy) {
+  clearPreviousSyncedHead(childWindow)
+  const styleLoadedPromises: Array<Promise<any>> = []
```
khufu-dev: Avoid `any`, ideally there's a strict type otherwise use `unknown`

```suggestion
  const styleLoadedPromises = []
```

packages/react/src/utils/windowStyleSync.ts
```diff
@@ -0,0 +1,77 @@
...
+  for (let i = 0; i < document.head.children.length; i++) {
+    const n = document.head.children[i].cloneNode(true) as any
```
khufu-dev: Why is it casting to `any`? It essentially disables all the type checking below.
```suggestion
    const n = document.head.children[i].cloneNode(true)
```

packages/react/src/spatialized-container/SpatializedContainer.tsx
```diff
@@ -12,35 +12,64 @@ import { PortalInstanceContext } from './context/PortalInstanceContext'
...
+    onSpatialRotateEnd: _6,
+    onSpatialMagnify: _7,
+    onSpatialMagnifyEnd: _8,
+    spatializedContent: _content,
+    createSpatializedElement: _create,
+    getExtraSpatializedElementProperties: _getExtra,
+    extraRefProps: _extraRef,
+    sizingMode: _sizingMode,
+    ...restProps
+  } = inprops as any
```
khufu-dev: Please fix cast to `any`

packages/react/src/spatialized-container/SpatializedContainer.tsx
```diff
@@ -12,35 +12,64 @@ import { PortalInstanceContext } from './context/PortalInstanceContext'
...
+  innerRef,
+  ...inprops
+}: SpatializedContainerProps<T> & {
+  innerRef: ForwardedRef<SpatializedElementRef<T>>
+}) {
+  const {
+    component: Component,
+    children,
+    'enable-xr': _enableXR,
+    onSpatialTap: _1,
```
khufu-dev: Why can't we keep the original name as unused variable like before?

packages/core/src/reality/Attachment.ts
```diff
@@ -0,0 +1,45 @@
...
+  constructor(
+    id: string,
+    private readonly windowProxy: WindowProxy,
+    private options: AttachmentEntityOptions,
+  ) {
+    super(id)
+  }
+
+  getContainer(): HTMLElement {
+    return (this.windowProxy as any).document.body
```
khufu-dev: Please fix cast to `any`

packages/react/src/reality/components/AttachmentEntity.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+export const AttachmentEntity: React.FC<AttachmentEntityProps> = ({
+  attachment: attachmentName,
+  position,
+  size,
+}) => {
+  const ctx = useRealityContext()
+  const parent = useParentContext()
+  const attachmentRef = useRef<Attachment | null>(null)
+  const parentIdRef = useRef<string | null>(null)
```
khufu-dev, line 26: Null is inferred
```suggestion
  const parentIdRef = useRef<string>(null)
```

packages/react/src/reality/components/AttachmentEntity.tsx
```diff
@@ -0,0 +1,152 @@
...
+}
+
+export const AttachmentEntity: React.FC<AttachmentEntityProps> = ({
+  attachment: attachmentName,
+  position,
+  size,
+}) => {
+  const ctx = useRealityContext()
+  const parent = useParentContext()
+  const attachmentRef = useRef<Attachment | null>(null)
```
khufu-dev, line 25:
```suggestion
  const attachmentRef = useRef<Attachment>(null)
```

packages/react/src/reality/components/AttachmentEntity.tsx
```diff
@@ -0,0 +1,152 @@
...
+          att.destroy()
+          return
+        }
+        // Initial style sync for attachment window
+        const windowProxy = att.getWindowProxy()
+        setOpenWindowStyle(windowProxy)
+        windowProxy.document.body.style.display = 'block'
+        windowProxy.document.body.style.minWidth = '100%'
+        windowProxy.document.body.style.maxWidth = '100%'
+        windowProxy.document.body.style.minHeight = '100%'
```
khufu-dev, line 60: Aren't these the default styles for body?

packages/react/src/reality/components/AttachmentEntity.tsx
```diff
@@ -0,0 +1,152 @@
...
+        if (!viewport) {
+          const meta = windowProxy.document.createElement('meta')
+          meta.name = 'viewport'
+          meta.content =
+            'width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no'
+          windowProxy.document.head.appendChild(meta)
+        }
+
+        // Ensure base href for relative URLs
+        const base = windowProxy.document.createElement('base')
```
khufu-dev, line 76: That's cool didn't know about this feature.

packages/visionOS/web-spatial/webview/SpatialWebController.swift
```diff
@@ -73,9 +73,12 @@ class SpatialWebController: NSObject, WKNavigationDelegate, WKScriptMessageHandl
             firstLoad = false
         }
         var needAllow = deciside ?? false
-        
+
         if !needAllow{
-            UIApplication.shared.open(navigationAction.request.url!, options: [:], completionHandler: nil)
+            // Don't attempt to open internal webspatial:// protocol URLs externally
+            if navigationAction.request.url?.scheme != "webspatial" {
```
khufu-dev: Is this required for attachments?

packages/visionOS/web-spatial/webview/SpatialWebViewModel.swift
```diff
@@ -76,6 +76,10 @@ class SpatialWebViewModel {
     }
 
     func getController() -> SpatialWebController {
+        if controller == nil {
+            controller = SpatialWebController()
+            controller!.model = self
```
khufu-dev, line 83: Note that if this class is used across multiple threads then `controller` can be set to nil between line 80 and 81.

packages/react/src/reality/components/AttachmentEntity.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  // Ongoing style sync when parent document head changes
+  useEffect(() => {
+    if (!childWindow) return
+    let timer: number | undefined
+    const scheduleSync = () => {
+      if (timer) window.clearTimeout(timer)
+      timer = window.setTimeout(() => {
+        syncParentHeadToChild(childWindow)
+      }, 100)
```
khufu-dev, line 136: Document why there's a timeout.

packages/react/src/spatialized-container/SpatializedContainer.tsx
parthpatelsj: Reverted to keep original name

packages/react/src/reality/components/AttachmentEntity.tsx
parthpatelsj: setOpenWindowStyle() is called on the line above, which sets display: inline-block, minWidth/maxWidth: fit-content, etc. (to shrink wrap the content). 

Attachments need the opposite: the body should fill the attachment frame set by RealityKit. These lines override those styles to display: block and 100% dimensions. Updated the comment to clarify this.

packages/visionOS/web-spatial/webview/SpatialWebController.swift
parthpatelsj: The guard isn't specific to attachments — it's a general safety net. webspatial:// is an internal scheme that normally goes through the open window delegate, not the navigation delegate. 

But there are paths where a webspatial:// URL can hit the navigation delegate. In those cases needAllow ends up false and the URL would get handed to UIApplication.shared.open without this guard. It was added during attachment work but applies to both scene and attachment webviews. 

Found this when I was developing test for attachments in test-server

packages/visionOS/web-spatial/webview/SpatialWebViewModel.swift
parthpatelsj: Ack. 

This is currently main-thread only, but the nil check + force unwrap pattern is inherently unsafe if that changes. There's a broader issue here —Filed a follow-up to separate the attachment webview lifecycle so this fallback isn't needed at all. (See comment in AttachmentManager.swift)

packages/react/src/reality/components/AttachmentEntity.tsx
parthpatelsj: useRef<T>(null) in React 18 types makes .current read-only. We need it mutable — the attachment doesn't exist yet at render time, so the ref starts as null, gets assigned after async creation in useEffect, then gets nulled out again on cleanup. 

so tsc errors out without | null. I think React 19 types fixes this, I added a comment.

PR #1010, opened 2026-03-06
fix(react): prevent Model ref from being used before element is created

### LINKED ISSUE #1001: Calling entityTransform or ready on static 3D model causes crash
A crash could occur due to calling `ref.current.ready` or `ref.current.entityTransform` before the spatialized static 3D model element is created.  

**Rootcause**
1. When a static 3D model is created it calls the async [`createSpatializedStatic3DElement` function](https://github.com/webspatial/webspatial-sdk/blob/e916bcfb489dba4d0847da1c4c7ab2dec7a65b07/packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx#L114) which calls a JSB function to create the native static 3D model. 
2. The extra DOM APIs on static 3D model such as [`ready` and `entityTransform`](https://github.com/webspatial/webspatial-sdk/blob/e916bcfb489dba4d0847da1c4c7ab2dec7a65b07/packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx#L129) require [`__spatializedElement`](https://github.com/webspatial/webspatial-sdk/blob/e916bcfb489dba4d0847da1c4c7ab2dec7a65b07/packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx#L131) to be set on the DOM entity.  
3. But `__spatializedElement` is not set until the async `createSpatializedStatic3DElement` function returns. This results in a crash if `ready` or `entityTransform` is called before `createSpatializedStatic3DElement` completes

khufu-dev:
Fix #1001. Wait until createSpatializedElement has completed before setting ref on Model component.

Simulator Screenshot - Apple Vision Pro 26 2 - 2026-03-05 at 19 59 45

### DIFF
--- a/packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx
+++ b/packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx
@@ -5,6 +5,9 @@ import {
   useContext,
   useEffect,
   useMemo,
+  useState,
+  useImperativeHandle,
+  useRef,
 } from 'react'
 import { SpatializedContainer } from './SpatializedContainer'
 import { getSession } from '../utils'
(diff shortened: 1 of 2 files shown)


packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx
```diff
@@ -110,14 +113,22 @@ function SpatializedContent(props: SpatializedStatic3DContentProps) {
...
 function SpatializedStatic3DElementContainerBase(
   props: SpatializedStatic3DContainerProps,
   ref: ForwardedRef<SpatializedStatic3DElementRef>,
 ) {
+  const containerRef = useRef<SpatializedStatic3DElementRef>(null)
+  const [elementCreated, setElementCreated] = useState(false)
+  useImperativeHandle<
+    SpatializedStatic3DElementRef | null,
+    SpatializedStatic3DElementRef | null
+  >(ref, () => (elementCreated ? containerRef.current : null), [elementCreated])
```
parthpatelsj, line 125: With this, on initial mount developers cannot do `await ref.current.ready` inside a useEffect on mount since it would null, right? They would have to figure out when ref is ready. 

Example: 
```const myModelRef = useRef<SpatializedStatic3DElementRef>(null);

useEffect(() => {
  // `myModelRef.current` is null on mount.
  myModelRef.current?.ready.then(() => {
    console.log("Model is fully loaded and ready!");
  });
}, []);```

Currently ^ this example will cause JS error. Maybe we can document if this is by design.

**Review by parthpatelsj — APPROVED** [review]

File: packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx (hunk 1 above)
Comment (khufu-dev) [reply packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx:125]: That code should work, no JS error will happen. You just need to listen to ref updates in the dependency like any normal React ref otherwise the promise will never execute.

```jsx
useEffect(() => {
  ref.current?.ready.then(() => console.log("Model is ready!"));
}, [ref.current]);
```

PR #1014, opened 2026-03-06
fix(visionOS): cleanup scene on SpatialNavView navigation

### LINKED ISSUE #1008: navigating [SpatialNavView] does not clean up
Navigating within a WebSpatial app using SpatialNavView (back, forward, reload, or navigating to a new page) does not properly clean up spatialized elements from the previous page. Old SpatialDiv panels and other spatial objects remain visible in the scene after the page changes, overlapping with the new page's content.

This is visible in test-server

Image
Image

Open a WebSpatial app in the App Shell
Navigate to a page that creates spatialized elements (e.g., a page with enable-xr divs)
Using SpatialNavView, either:

Reload the page (refresh button), or
Navigate to a new page, then press back, then forward

Observe that the spatialized elements from the previous page load persist — they are not destroyed before the new page's elements are created

Expected: On every navigation event (reload, back, forward, new URL), all existing SpatialObjects in the current SpatialScene should be destroyed before the new page begins rendering spatial content.
Actual: Old spatialized elements accumulate. The scene ends up with duplicate/stale panels from prior navigations.

walton007:
Fixes #1008\n\nSummary\n- Clean up current SpatialScene before SpatialNavView back/forward/reload/URL navigation to prevent stale SpatialDiv/panels accumulating across SPA navigations.\n- Add a visionOS unit test for navigation cleanup.\n- Fix visionOS test host settings so xcodebuild tests run on the simulator.\n\nNotes\n- Changeset added for @webspatial/platform-visionos (patch).\n


packages/visionOS/web-spatial/view/SpatialNavView.swift
```diff
@@ -126,13 +126,15 @@ struct SpatialNavView: View {
...
+                        )
                         .disabled(!(self.canGoBack))
                         NavButton(
                             action: { self.goForward()
                             },
                             children: Image("arrow_right"),
-                            clearBackGround: true)
+                            clearBackGround: true
+                        )
                         .disabled(!(self.canGoBack))
```
parthpatelsj, line 140: should be canGoForward?

parthpatelsj (approved):

packages/visionOS/web-spatial/view/SpatialNavView.swift
```diff
@@ -187,14 +191,14 @@ struct SpatialNavView: View {
...
-                    .background(.black)
-                    .cornerRadius(100)
+                .lineLimit(1)
+                .textSelection(.enabled)
+                .padding(12)
+                .frame(minWidth: 200)
+                .frame(maxWidth: 500)
+                .frame(height: 44)
+                .background(.black)
+                .cornerRadius(100)
```
khufu-dev: There are a lot of formatting and cleanup changes in this pull request unrelated to the actual fix. Can you split those into a separate pull request so that we can pick the fix onto stable branch?

packages/visionOS/web-spatial/view/SpatialNavView.swift
parthpatelsj: *this is not related to this PR technically

packages/visionOS/web-spatial/view/SpatialNavView.swift
walton007: Yes, auto format is applied. Need to be seperated.

PR #1032, opened 2026-03-13
fix(react): stabilize head style sync for spatial windows

### LINKED ISSUE #1033: Fix unstable head style sync in spatial windows and guard non-string className
## Overview
This issue tracks two related stability problems that were observed while using spatial windows (SpatialDiv / attachments) in the React SDK.

### Symptoms
When the host page `<head>` changes rapidly (e.g. Vite HMR or frequent dynamic `<style>` injections), the child window created for SpatialDiv/attachments may:
- accumulate duplicated synced stylesheet `<link>` elements over time, or
- have existing synced stylesheet `<link>` elements removed/replaced during rapid updates.

### Expected
- Synced stylesheet `<link>` elements in the child window remain stable (not replaced) when the host’s stylesheet list is unchanged.
- No duplicated synced stylesheet links are appended.

### Actual
- Duplicated stylesheet links can appear, or synced links can be replaced during rapid host `<head>` mutations.

### How to Reproduce
1. Render a SpatialDiv (`<div enable-xr>`) or an attachment portal.
2. Continuously mutate the host `<head>` (e.g. add/remove a `<style>` tag at a high frequency).
3. Observe the child window `<head>` for `link[rel=\"stylesheet\"][data-webspatial-sync=\"1\"]` changes.

### Symptoms
`replaceToSpatialPrimitiveType` previously assumed `props.className` is a string and called `.split(' ')`. Some third-party libraries (e.g. `NavLink` from react-router) may pass `className` as a function, which can cause a runtime exception.

### Expected
- No runtime error when a component provides `className` as a function or other non-string value.

### Actual
- A non-string `className` can trigger an exception during the JSX transform path.

### How to Reproduce
1. Use a component that accepts `className` as a function (e.g. `NavLink`) in a context where the JSX transform checks `className`.
2. The previous `.split(' ')` call on a non-string value can throw.

## Fix
- Use the shared `useSyncHeadStyles` hook for ongoing head synchronization.
- Make `syncParentHeadToChild` idempotent and concurrency-safe to avoid duplicate injection and to drop stale in-flight syncs.
- Add a type guard in `replaceToSpatialPrimitiveType` so it only splits `className` when it is a string.

## Verification
- Test-server probe: open `/#/head-style-sync` and click `Run host head stress`. Expect the probe to stay `PASS` without link duplication.
- Confirm a `NavLink` (or equivalent) passing `className` as a function no longer causes a runtime error.

## References
- PR: https://github.com/webspatial/webspatial-sdk/pull/1032

walton007:
Fixes #1033

## Summary
Stabilize head style synchronization for spatial windows (SpatialDiv/attachments) and harden the JSX transform against non-string `className` values.

## Changes
- Use the shared `useSyncHeadStyles` hook for ongoing head updates in both the SpatialDiv (Spatialized2DElement) path and the attachment window path.
- Make `syncParentHeadToChild` idempotent and concurrency-safe to avoid re-adding the same stylesheet links and to drop stale in-flight syncs.
- Guard `replaceToSpatialPrimitiveType` so it only calls `.split(' ')` when `props.className` is a string, avoiding runtime errors when third-party libs (e.g. `NavLink`) pass a function.
- Add a test-server probe page to stress host `<head>` mutations and validate that synced stylesheet links remain stable (not replaced/duplicated).
- Add a Changeset for `@webspatial/react-sdk`.

## Verification
- Run `pnpm -C apps/test-server dev` and open `/#/head-style-sync`.
- Click `Run host head stress`.
- Expect the probe to stay `PASS`, with the baseline stylesheet link still connected and the synced link count not increasing.
- Render a `NavLink` (or any component that provides `className` as a function) in the app; confirm no runtime exception is thrown by the JSX transform.

## Notes
- CI uses `pnpm install --frozen-lockfile`, so the lockfile is kept in sync with `packages/react/package.json` changes.

### DIFF
--- a/apps/test-server/src/pages/head-style-sync/index.tsx
+++ b/apps/test-server/src/pages/head-style-sync/index.tsx
@@ -1,28 +1,26 @@
-import {useRef, useCallback, useState} from 'react'
+import { useRef, useCallback, useState, useEffect } from 'react'
 import { enableDebugTool } from '@webspatial/react-sdk'
 
 enableDebugTool()
 
 function HeadSyncProbe() {
-  const rootRef = React.useRef<HTMLDivElement | null>(null)
-  const getTargetDocument = React.useCallback(() => {
+  const rootRef = useRef<HTMLDivElement | null>(null)
+  const getTargetDocument = useCallback(() => {
     return rootRef.current?.ownerDocument ?? null
   }, [])
 
-  const [status, setStatus] = React.useState<'pending' | 'pass' | 'fail'>(
-    'pending',
-  )
-  const [message, setMessage] = React.useState<string>('')
-  const baselineLinkRef = React.useRef<HTMLLinkElement | null>(null)
-  const baselineCountRef = React.useRef<number>(0)
-  const [baselineHref, setBaselineHref] = React.useState<string>('')
-  const [currentInfo, setCurrentInfo] = React.useState<{
+  const [status, setStatus] = useState<'pending' | 'pass' | 'fail'>('pending')
+  const [message, setMessage] = useState<string>('')
+  const baselineLinkRef = useRef<HTMLLinkElement | null>(null)
+  const baselineCountRef = useRef<number>(0)
+  const [baselineHref, setBaselineHref] = useState<string>('')
+  const [currentInfo, setCurrentInfo] = useState<{
     linkCount: number
     firstHref: string
     baselineConnected: boolean
   }>({ linkCount: 0, firstHref: '', baselineConnected: false })
 
-  const collect = React.useCallback(() => {
+  const collect = useCallback(() => {
     const targetDocument = getTargetDocument()
     if (!targetDocument) {
       setCurrentInfo({ linkCount: 0, firstHref: '', baselineConnected: false })
@@ -48,7 +46,7 @@ function HeadSyncProbe() {
   }, [])
 
   // Capture a baseline synced stylesheet link node.
-  React.useEffect(() => {
+  useEffect(() => {
     const startedAt = Date.now()
     const timer = window.setInterval(() => {
       const { links, first } = collect()
@@ -81,7 +79,7 @@ function HeadSyncProbe() {
   }, [collect])
 
   // Watch head changes and ensure synced stylesheet links are not replaced.
-  React.useEffect(() => {
+  useEffect(() => {
     if (!baselineLinkRef.current) return
 
     const check = () => {
@@ -169,10 +167,10 @@ function HeadSyncProbe() {
 }
 
 export default function HeadStyleSyncPage() {
-  const [running, setRunning] = React.useState(false)
-  const [runCount, setRunCount] = React.useState(0)
+  const [running, setRunning] = useState(false)
+  const [runCount, setRunCount] = useState(0)
 
-  const mutateHostHeadOnce = React.useCallback(() => {
+  const mutateHostHeadOnce = useCallback(() => {
     const id = 'webspatial-head-sync-test-style'
     const prev = document.getElementById(id)
     if (prev) prev.parentNode?.removeChild(prev)
@@ -183,7 +181,7 @@ export default function HeadStyleSyncPage() {
     document.head.appendChild(style)
   }, [])
 
-  const runStress = React.useCallback(async () => {
+  const runStress = useCallback(async () => {
     if (running) return
     setRunning(true)
     setRunCount(0)
--- a/packages/react/src/reality/components/AttachmentEntity.tsx
+++ b/packages/react/src/reality/components/AttachmentEntity.tsx
@@ -6,7 +6,7 @@ import {
   setOpenWindowStyle,
   syncParentHeadToChild,
 } from '../../utils/windowStyleSync'
-import { useSyncHeadStyles } from '../../utils/use-sync-head-styles'
+import { useSyncHeadStyles } from '../../utils/useSyncHeadStyles'
 
 let instanceCounter = 0
 
@@ -126,7 +126,7 @@ export const AttachmentEntity: React.FC<AttachmentEntityProps> = ({
     }
   }, [ctx, attachmentName])
 
-  useSyncHeadStyles(childWindow, { delayMs: 100, subtree: false })
+  useSyncHeadStyles(childWindow, { subtree: false })
 
   // Update position/size when they change
   useEffect(() => {
--- a/packages/react/src/spatialized-container/Spatialized2DElementContainer.tsx
+++ b/packages/react/src/spatialized-container/Spatialized2DElementContainer.tsx
@@ -14,7 +14,7 @@ import {
   setOpenWindowStyle,
   syncParentHeadToChild,
 } from '../utils/windowStyleSync'
-import { useSyncHeadStyles } from '../utils/use-sync-head-styles'
+import { useSyncHeadStyles } from '../utils/useSyncHeadStyles'
 import { getInheritedStyleProps, parseCornerRadius } from './utils'
 import {
   SpatialCustomStyleVars,
@@ -86,7 +86,6 @@ function SpatializedContent<P extends ElementType>(
   const { windowProxy } = spatialized2DElement
 
   useSyncHeadStyles(windowProxy, {
-    delayMs: 100,
     subtree: false,
   })
 
--- a/packages/react/src/utils/useSyncHeadStyles.ts
+++ b/packages/react/src/utils/useSyncHeadStyles.ts
@@ -3,7 +3,6 @@ import { useEffect } from 'react'
 import { syncParentHeadToChild } from './windowStyleSync'
 
 interface Options {
-  delayMs?: number
   subtree?: boolean
   immediate?: boolean
 }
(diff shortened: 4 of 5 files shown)


apps/test-server/src/pages/head-style-sync/index.tsx
```diff
@@ -0,0 +1,252 @@
+import React from 'react'
```
khufu-dev:
```suggestion
import {useRef, useCallback, useState} from 'react'
```

apps/test-server/src/pages/head-style-sync/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+import { enableDebugTool } from '@webspatial/react-sdk'
+
+enableDebugTool()
+
+function HeadSyncProbe() {
+  const rootRef = React.useRef<HTMLDivElement | null>(null)
```
khufu-dev:
```suggestion
  const rootRef = useRef<HTMLDivElement | null>(null)
```

packages/react/src/reality/components/AttachmentEntity.tsx
```diff
@@ -1,14 +1,17 @@
 import React, { useEffect, useRef, useState } from 'react'
+
```
khufu-dev:
```suggestion
```

packages/react/src/reality/components/AttachmentEntity.tsx
```diff
@@ -1,14 +1,17 @@
...
 import { Attachment } from '@webspatial/core-sdk'
+
```
khufu-dev, line 3:
```suggestion
```

packages/react/src/spatialized-container/Spatialized2DElementContainer.tsx
```diff
@@ -75,26 +78,17 @@ function useSyncDocumentTitle(
...
 function SpatializedContent<P extends ElementType>(
   props: SpatializedContentProps<SpatializedElementRef, P>,
 ) {
   const { spatializedElement, ...restProps } = props
   const spatialized2DElement = spatializedElement as Spatialized2DElement
-  const windowProxy = spatialized2DElement.windowProxy
+  const { windowProxy } = spatialized2DElement
 
-  useSyncHeaderStyle(windowProxy)
+  useSyncHeadStyles(windowProxy, {
```
khufu-dev, line 88: Good name change

packages/react/src/utils/useSyncHeadStyles.ts
```diff
@@ -0,0 +1,63 @@
+import { useEffect } from 'react'
```
khufu-dev, line 1: Rename the file to `useSyncHeadStyles.ts` to be consistent with React hook file naming

packages/react/src/utils/useSyncHeadStyles.ts
```diff
@@ -0,0 +1,63 @@
...
+interface Options {
+  delayMs?: number
+  subtree?: boolean
+  immediate?: boolean
+}
+
+function defaultShouldSync(mutations?: MutationRecord[] | null) {
+  if (!Array.isArray(mutations) || mutations.length === 0) return false
+  for (const mutation of mutations) {
+    const nodes: Node[] = [
```
khufu-dev, line 13:
```suggestion
    const nodes = [
```

packages/react/src/utils/useSyncHeadStyles.ts
```diff
@@ -0,0 +1,63 @@
...
+      ...Array.from(mutation.addedNodes),
+      ...Array.from(mutation.removedNodes),
+    ]
+    for (const node of nodes) {
+      if (!(node instanceof Element)) continue
+      const tag = node.tagName
+      if (tag === 'STYLE') return true
+      if (tag === 'LINK') {
```
khufu-dev, line 21: According to [W3C specs](https://html.spec.whatwg.org/multipage/semantics.html#the-link-element) the link tag can be in the head or body. Do we support both?

packages/react/src/utils/use-sync-head-styles.ts
```diff
@@ -0,0 +1,63 @@
+import { useEffect } from 'react'
+
+import { syncParentHeadToChild } from './windowStyleSync'
+
+interface Options {
+  delayMs?: number
```
khufu-dev: Should we omit this property so we can standardize the timeout.

packages/react/src/utils/windowStyleSync.ts
```diff
@@ -1,36 +1,44 @@
...
-    // ~1/50 loads, if the same url is loaded very quickly in a window and a child window,
-    // the second load request never is fired resulting in css not to be applied.
-    // Workaround this by making the css stylesheet request unique
-    n.href += '?uniqueURL=' + Math.random()
-    n.onerror = function (error) {
-      console.error('Failed to load style link', (n as HTMLLinkElement).href)
-      resolve(false)
+    const { href } = n
+    const sep = href.includes('?') ? '&' : '?'
+    n.href = `${href}${sep}uniqueURL=${Math.random()}`
```
khufu-dev:
```suggestion
    n.href = `${href}${sep}uniqueURL=${crypto.randomUUID()}`
```

packages/react/src/utils/windowStyleSync.ts
```diff
@@ -1,36 +1,44 @@
...
+    let finished = false
+    const finish = (ok: boolean) => {
+      if (finished) return
+      finished = true
+      resolve(ok)
+    }
+
+    n.onerror = () => {
+      finish(false)
+    }
```
khufu-dev:
```suggestion
    n.onerror = () =>  finish(false)
```

packages/react/src/utils/windowStyleSync.ts
```diff
@@ -1,36 +1,44 @@
 export function asyncLoadStyleToChildWindow(
   childWindow: WindowProxy,
   n: HTMLLinkElement,
```
khufu-dev: `n` is not descriptive enough.
```suggestion
  link: HTMLLinkElement,
```

packages/react/src/utils/windowStyleSync.ts
```diff
@@ -1,36 +1,44 @@
...
-    n.onload = () => resolve(true)
 
-    // need to wait for some time to make sure the style is loaded
-    // otherwise, the style may not be applied
     setTimeout(() => {
+      if (!isCurrent()) {
+        finish(false)
+        return
+      }
       childWindow.document.head.appendChild(n)
```
khufu-dev: Will this throw an error if childWindow is closed before the timeout completes?

packages/react/src/utils/windowStyleSync.ts
```diff
@@ -1,36 +1,44 @@
...
+        n.parentNode?.removeChild(n)
+        finish(false)
+        return
+      }
+      finish(true)
     }
-    n.onload = () => resolve(true)
 
-    // need to wait for some time to make sure the style is loaded
-    // otherwise, the style may not be applied
```
khufu-dev, line 18: Keep the comment to justify the timeout.

packages/react/src/spatialized-container/Spatialized2DElementContainer.tsx
walton007: What's your suggest name?

packages/react/src/utils/use-sync-head-styles.ts
walton007: OK

packages/react/src/utils/useSyncHeadStyles.ts
walton007: I hope keep the type declaration

packages/react/src/utils/useSyncHeadStyles.ts
walton007: No. We support head only now.

packages/react/src/utils/windowStyleSync.ts
walton007: OK

packages/react/src/utils/windowStyleSync.ts
walton007: Prefer Math.random, which is faster.

packages/react/src/utils/windowStyleSync.ts
walton007: Can keep this coding style, which might help adding console info

packages/react/src/utils/windowStyleSync.ts
walton007: I think isCurrent function is used to avoid this

packages/react/src/utils/windowStyleSync.ts
walton007: Agree. Keep the comment

packages/react/src/spatialized-container/Spatialized2DElementContainer.tsx
walton007: We can have another PR when you have a better name candidate

khufu-dev: Sorry, I meant to communicate that this is a good name change 👍

PR #1034, opened 2026-03-13
chore: improve entity type usage & test page

mzvast:
clean up outdated test page. improve entity type declaration.

### DIFF
--- a/packages/core/src/types/types.ts
+++ b/packages/core/src/types/types.ts
@@ -115,6 +115,16 @@ export interface SpatialSceneCreationOptions {
   baseplateVisibility?: BaseplateVisibilityType
 }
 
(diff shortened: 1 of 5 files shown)


packages/core/src/types/types.ts
```diff
@@ -186,8 +186,6 @@ export interface SpatialEntityProperties {
   scale: Vec3
 }
 
-export type SpatialEntityEventType = 'spatialtap' //| 'drag' | 'rotate' | 'scale'
```
khufu-dev, line 189: Wouldn't it be better to keep this type so that developer knows what is allowed?

mzvast: correct, keep it and extend with new type

PR #1042, opened 2026-03-17
add createElement for old jsx transform

mzvast:
in certain case, transpiler will fallback to old jsx transform. There is no way to disable this fallback, except changing user code.

example 
```
const TestPage = forwardRef((props, ref) => {
  const Comps:any[] = [{Comp:()=><div>comp1</div>,withRef:true},{Comp:()=><div>comp2</div>,withRef:true}]

  return (
    <>
    <div enable-xr > testdiv</div>
      {Comps.map(({ Comp }, index) =>
        <Comp {...props}  key={index} />
      )}
    </>
  )
})
```

this wil generate like this
```
const TestPage = reactExports.forwardRef((props, ref) => {
  const Comps = [{ Comp: () => /* @__PURE__ */ jsx("div", { children: "comp1" }), withRef: true }, { Comp: () => /* @__PURE__ */ jsx("div", { children: "comp2" }), withRef: true }];
  return /* @__PURE__ */ jsxs(jsxRuntimeExports.Fragment, { children: [
    /* @__PURE__ */ jsx("div", { "enable-xr": true, children: " testdiv" }),
    Comps.map(
      ({ Comp }, index2) => /* @__PURE__ */ createElement(Comp, { ...props, key: index2 })
    )
  ] });
});
```

in which `createElement` is imported from `@webspatial/react-sdk`


packages/react/src/jsx/jsx-shared.ts
```diff
@@ -69,3 +70,12 @@ export function jsxDEV(
...
 }
+
+export function createElement(
+  type: React.ElementType,
+  props?: unknown,
+  ...children: any[]
+) {
+  type = replaceToSpatialPrimitiveType(type, props)
+  return reactCreateElement(type, props as any, ...children)
+}
```
khufu-dev: Can you type this properly so that it uses the same types as React `createElement`. When we upgrade React the typing should automatically reflect the new types.
```suggestion
export function createElement(...args: Parameters<typeof reactCreateElement>) {
  const [type, props, ..rest] = args
  const newType = replaceToSpatialPrimitiveType(type, props)
  return reactCreateElement(newType, props, ...rest)
}
```

.changeset/puny-lines-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+'@webspatial/react-sdk': patch
```
khufu-dev: Introducing a new JSX API should at least be a minor release possibly even a major release.

mzvast: ok, changed to minor

packages/react/src/jsx/jsx-shared.ts
```diff
@@ -69,3 +70,9 @@ export function jsxDEV(
   type = replaceToSpatialPrimitiveType(type, props)
   return _jsxDEV(type, props, key, isStatic, source, self)
 }
+
+export function createElement(...args: Parameters<typeof reactCreateElement>) {
+  const [type, props, ...rest] = args
+  const newType = replaceToSpatialPrimitiveType(type as any, props)
```
khufu-dev, line 76:
```suggestion
  const newType = replaceToSpatialPrimitiveType(type, props)
```

khufu-dev (approved):

packages/react/src/jsx/jsx-shared.ts
mzvast: `type` comes from `Parameters<typeof React.createElement>` (an overload union). When we destructure it, TS widens the tag case to plain `string`, which no longer matches the stricter `keyof JSX.IntrinsicElements` overload and causes a false-positive error. `as any` is a localized escape hatch for this overload inference limitation; runtime behavior is unchanged and React still validates the element type.

PR #1044, opened 2026-03-17
Attachment JSB Initialization & Ref Cleanup

parthpatelsj:
- Reliable Attachment Init: Moves metadata delivery (position, size, parent) from URL query parameters to a dedicated InitializeAttachmentCommand JSB. This ensures the native side (SpatialScene.swift) matches metadata to the correct WKWebView via a stable ID, avoiding URL parsing races.

- Ref-Based Cleanup: Refactors useSpatializedElement and AttachmentEntity to store native instances in useRef. This ensures the .destroy() call in the useEffect cleanup consistently targets the actual active instance, preventing memory leaks and orphaned native resources.

### DIFF
--- a/packages/react/src/reality/components/AttachmentEntity.tsx
+++ b/packages/react/src/reality/components/AttachmentEntity.tsx
@@ -2,10 +2,8 @@ import React, { useEffect, useRef, useState } from 'react'
 import { Attachment } from '@webspatial/core-sdk'
 
 import { useRealityContext, useParentContext } from '../context'
-import {
-  setOpenWindowStyle,
-  syncParentHeadToChild,
-} from '../../utils/windowStyleSync'
+import { setOpenWindowStyle } from '../../utils/windowStyleSync'
+import { useSyncHeadStyles } from '../../utils/useSyncHeadStyles'
 
 let instanceCounter = 0
 
@@ -60,7 +58,6 @@ export const AttachmentEntity: React.FC<AttachmentEntityProps> = ({
         windowProxy.document.body.style.minWidth = '100%'
         windowProxy.document.body.style.maxWidth = '100%'
         windowProxy.document.body.style.minHeight = '100%'
-        await syncParentHeadToChild(windowProxy)
 
         // Ensure viewport meta
         const viewport = windowProxy.document.querySelector(
@@ -126,26 +123,7 @@ export const AttachmentEntity: React.FC<AttachmentEntityProps> = ({
     }
   }, [ctx, attachmentName])
 
-  // Ongoing style sync when parent document head changes
-  useEffect(() => {
-    if (!childWindow) return
-    let timer: number | undefined
-    const scheduleSync = () => {
-      if (timer) window.clearTimeout(timer)
-      timer = window.setTimeout(() => {
-        syncParentHeadToChild(childWindow)
-      }, 100)
-    }
-
-    scheduleSync()
-
-    const observer = new MutationObserver(scheduleSync)
-    observer.observe(document.head, { childList: true, subtree: true })
-    return () => {
-      if (timer) window.clearTimeout(timer)
-      observer.disconnect()
-    }
-  }, [childWindow])
+  useSyncHeadStyles(childWindow, { subtree: false })
 
   // Update position/size when they change
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


packages/react/src/reality/components/AttachmentEntity.tsx
```diff
@@ -126,7 +126,26 @@ export const AttachmentEntity: React.FC<AttachmentEntityProps> = ({
     }
   }, [ctx, attachmentName])
 
-  useSyncHeadStyles(childWindow, { subtree: false })
+  // Ongoing style sync when parent document head changes
+  useEffect(() => {
```
khufu-dev: Why not re-use the existing hook? This is very similar code.

packages/react/src/spatialized-container/hooks/useSpatializedElement.ts
```diff
@@ -8,17 +8,18 @@ export function useSpatializedElement(
...
     useState<SpatializedElement>()
+  const elementRef = useRef<SpatializedElement | undefined>(undefined)
 
   useEffect(() => {
     let isDestroyed = false
-    let spatializedElement: SpatializedElement | undefined
+
     createSpatializedElement().then(
       (inSpatializedElement: SpatializedElement) => {
+        if (!inSpatializedElement) return
```
khufu-dev, line 24: Why is this check necessary? `inSpatializedElement` should not be null/undefined.

packages/react/src/spatialized-container/hooks/useSpatializedElement.ts
```diff
@@ -8,17 +8,18 @@ export function useSpatializedElement(
 ) {
   const [spatializedElement, setSpatializedElement] =
     useState<SpatializedElement>()
+  const elementRef = useRef<SpatializedElement | undefined>(undefined)
```
khufu-dev, line 16: I don't understand this change. Why are we storing the element in two variables? Wouldn't the correct fix be to add `spatializedElement` to the dependency array of the `useEffect` hook.

packages/react/src/reality/components/AttachmentEntity.tsx
parthpatelsj: You're right — useSyncHeadStyles already accepts WindowProxy and handles this. Reverted to the hook call

packages/react/src/spatialized-container/hooks/useSpatializedElement.ts
parthpatelsj: This is a very good callout/question. 

**State**: we need the hook to return the element and trigger a re-render when it’s ready. If we only had a ref, we’d never re-render and callers would never get the element. We can’t put that state in the effect deps or we get: create → setState → deps change → effect re-runs → create again → infinite loop.

**Ref**: the element is created asynchronously in .then(). The effect returns its cleanup synchronously, so when cleanup runs (unmount or deps change) the promise may not have resolved yet. Cleanup still needs to destroy the element if it was ever set — so we need a mutable value that both the .then() callback writes to and the cleanup reads from. A local variable inside the effect would only be set when .then() runs, so cleanup could run too early and miss it. The ref is that shared storage.

packages/react/src/spatialized-container/hooks/useSpatializedElement.ts
parthpatelsj: useSpatializedElement doesn't own creation — it calls whatever createSpatializedElement the caller passes in. 

Reality.tsx passes createReality, which resolves to null on cancellation or failure. The hook's .then() receives that null. 

Without the guard, we'd pass null into attachSpatializedElement and store it in state.
The type was lying — said Promise<SpatializedElement> but null is a valid resolved value. Updated to Promise<SpatializedElement | null> so it's more clear.

khufu-dev: Careful, this is happening because there's a `ts-ignore` in [Reality.tsx](https://github.com/webspatial/webspatial-sdk/blob/main/packages/react/src/reality/components/Reality.tsx#L121) which is very dangerous and should be fixed instead of specifying the wrong type here. The rest of the type system will not assume a null value.

packages/react/src/spatialized-container/hooks/useSpatializedElement.ts
khufu-dev: What if you separated the creation and cleanup into two separate `useEffects` can you avoid the infinite loop and use just the state?

PR #1045, opened 2026-03-18
feat: add dynamic 3D runtime updates

parthpatelsj:
### Summary
This PR introduces runtime reactivity to our 3D API. Previously, entities were mostly static after instantiation. We can now dynamically update material properties, modify geometry in-place, and apply material overrides to model entities on the fly.

### Architectural Overview
To support runtime updates without tanking performance or losing entity state (transforms, event listeners, hierarchy), this PR introduces three distinct update strategies depending on what is changing:

1. Material Reactivity (State Propagation)
Instead of recreating entities when a color or opacity changes, we now mutate the material and propagate that change.

React Layer: UnlitMaterial now watches its props. On change, it triggers an update.

Native Bridge: Added a new UpdateUnlitMaterialProperties JSB command.

Swift Layer: The Swift material updates its internal state (currentColor, currentOpacity, etc.) and automatically forces a refresh on any ModelComponent currently referencing it. Partial updates (e.g., tweaking just opacity) are fully supported.

2. Geometry Modification (Component Swapping)
Changing geometry (width, radius, etc.) requires rebuilding the mesh, but we don't want to destroy the underlying SpatialEntity and lose its position or event listeners.

Diffing: Introduced shallowEqual utilities to detect actual geometry prop changes.

In-Place Swap: When geometry changes, we simply remove the old ModelComponent from the entity and attach a newly generated one. The host entity remains intact.

3. Model Entity Updates & Overrides (Full Recreation / Hierarchy Traversal)
Material Overrides: Added a materials prop to ModelEntity. Under the hood, the new SetMaterialsOnEntity command walks the native entity hierarchy and injects the new materials into all relevant ModelComponent instances.

Entity Tear-down (recreateKey): For total model asset swaps, we need a clean slate. Added a recreateKey prop to useEntity/BaseEntity. When this key changes, it explicitly tears down the old entity and rebuilds it from scratch (resetting transforms/events as expected).

State Tracking: Take a look at SpatialComponent.swift and SpatialMaterial.swift. I had to add several properties to track mutable state (like spatialMaterials and mesh) so we have the context needed to rebuild components when the React tree pushes updates.

Backward Compatibility: All of these changes gracefully fall back to the existing static creation paths. No breaking changes to existing static implementations.

Before:
https://github.com/user-attachments/assets/60e99920-aced-4f3c-b0ad-4ce82557a36d

After:
https://github.com/user-attachments/assets/c00de7ce-b504-4783-a5a4-1637812a4262

### DIFF
--- a/packages/react/src/reality/components/GeometryEntity.tsx
+++ b/packages/react/src/reality/components/GeometryEntity.tsx
@@ -73,6 +73,10 @@ export const GeometryEntity = forwardRef<EntityRefShape, GeometryEntityProps>(
         try {
           const oldComponent = componentRef.current
           if (oldComponent) {
+            // TODO(P2): Rebuild removes/destroys the model component but not the previous mesh
+            // `SpatialGeometry`; those objects stay registered until explicitly destroyed (native
+            // `SpatialModelComponent.onDestroy` also only nils `mesh`). Destroy the prior mesh after
+            // teardown or fix cascade destroy on the Swift side to cap memory on long-lived updates.
             await entity.removeComponent(oldComponent)
             await oldComponent.destroy()
             componentRef.current = null
--- a/packages/react/src/reality/components/ModelEntity.tsx
+++ b/packages/react/src/reality/components/ModelEntity.tsx
@@ -22,9 +22,9 @@ export const ModelEntity = forwardRef<EntityRefShape, Props>(
     const entityRef = useRef<CoreSpatialModelEntity | null>(null)
     const lastMaterialsRef = useRef<string[] | undefined>(undefined)
 
-    // TODO: Overlapping async `setMaterials` calls (e.g. rapid `materials` prop changes) can apply
-    // out of order, like GeometryEntity's rebuild race. Use a generation counter (see
-    // `rebuildGen` in GeometryEntity) so stale applies are ignored after awaits.
+    // TODO(P2): Material overrides run async (`Promise.all` + `setMaterials`) with no generation/abort
+    // guard; a slower resolve can apply after a newer prop update and leave native materials stale vs
+    // React. Mirror `rebuildGen` in GeometryEntity so only the latest apply runs after awaits.
     // Dynamic material override (including clearing: props may go from ids to undefined / [])
     useEffect(() => {
       if (!ctx || !entityRef.current) return
--- a/packages/visionOS/web-spatial/model/dynamic3d/SpatialComponent.swift
+++ b/packages/visionOS/web-spatial/model/dynamic3d/SpatialComponent.swift
@@ -89,6 +89,9 @@ class SpatialModelComponent: SpatialComponent {
     }
 
     override func onDestroy() {
+        // TODO(P2): `mesh` is a registered `Geometry` spatial object; clearing the reference does not
+        // run `Geometry.destroy()`, so dynamic mesh rebuilds can leak mesh/registry entries until the
+        // scene ends. Call `mesh?.destroy()` (or equivalent) before nil-ing when ownership is exclusive.
         _resource = nil
         spatialMaterials = []
         mesh = nil


packages/react/src/reality/components/GeometryEntity.tsx
```diff
@@ -18,6 +25,78 @@ export const GeometryEntity = forwardRef<EntityRefShape, GeometryEntityProps>(
...
+          })
+
+          // Add to entity
+          await entity.addComponent(modelComponent)
+          componentRef.current = modelComponent
+        } catch (error) {
+          console.error('GeometryEntity: rebuild failed', error)
+        }
+      }
+      rebuild()
```
mzvast, line 121: if props changed frequently, there may be racing conditions when several rebuilds running at the same time.

Do we need cancel check at every await point?

useEffect(() => {
  let cancelled = false
  const rebuild = async () => {
    // await point
    if (cancelled) return
    // ...
  }
  rebuild()
  return () => { cancelled = true }
}, [/* deps */])

packages/react/src/reality/components/ModelEntity.tsx
```diff
@@ -1,33 +1,90 @@
...
+            )
+          ).filter(Boolean)
+          if (entityRef.current && materialList.length > 0) {
+            await entityRef.current.setMaterials(materialList)
+          }
+        } catch (error) {
+          console.error('ModelEntity: failed to set materials', error)
+        }
+      }
+      apply()
```
mzvast, line 50: same as rebuild ,racing conditions may happen here

packages/react/src/reality/components/GeometryEntity.tsx
```diff
@@ -18,6 +25,78 @@ export const GeometryEntity = forwardRef<EntityRefShape, GeometryEntityProps>(
...
+    }, [
+      ctx,
+      geometryOptions?.width,
+      geometryOptions?.height,
+      geometryOptions?.depth,
+      geometryOptions?.cornerRadius,
+      geometryOptions?.splitFaces,
+      geometryOptions?.radius,
```
mzvast: can we use geometryOptions object as dependency?

packages/react/src/reality/components/UnlitMaterial.tsx
```diff
@@ -29,8 +32,23 @@ export const UnlitMaterial: React.FC<Props> = ({ children, ...options }) => {
...
       // Use registry to schedule destruction after promise resolves
       resourceRegistry.removeAndDestroy(options.id)
+      materialRef.current = undefined
+      isInitializedRef.current = false
     }
   }, [ctx])
 
+  // Dynamic property updates
+  useEffect(() => {
+    if (!isInitializedRef.current || !materialRef.current) return
```
mzvast, line 46: After isInitializedRef.current = true, this effect will run and call updateProperties using the same options as Line20?
Should we do diff between new values and old one here?

packages/react/src/reality/components/GeometryEntity.tsx
parthpatelsj: I have added a generation counter `rebuildGen`

Each time props require rebuild, I increment `rebuildGen` After each await, I check `gen !== mutableRef.current.rebuildGen`. If this is true, then async is outdated, and stop/cleanup required.

packages/react/src/reality/components/ModelEntity.tsx
parthpatelsj: Yes, good point, I will need to revise to do similar solution as GeometryEntity

packages/react/src/reality/components/UnlitMaterial.tsx
parthpatelsj: setting isInitializedRef.current = true does not trigger a re-render/effect run by itself since isInitializedRef is a ref not state

packages/visionOS/web-spatial/model/dynamic3d/SpatialModelEntity.swift
```diff
@@ -15,12 +18,41 @@ class SpatialModelEntity: SpatialEntity {
...
+
+    func usesMaterial(_ materialId: String) -> Bool {
+        overrideSpatialMaterials.contains { $0.id == materialId }
+    }
+
+    private func applyOverrideMaterials() {
+        guard let modelEntity = modelEntity else { return }
+        func applyMaterials(to entity: Entity) {
+            if var modelComp = entity.components[ModelComponent.self] {
+                modelComp.materials = overrideSpatialMaterials.compactMap { $0.resource }
```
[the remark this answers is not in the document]
parthpatelsj: I have added this as a todo, and will take this as a follow up PR.

packages/react/src/reality/components/GeometryEntity.tsx
```diff
@@ -1,22 +1,122 @@
...
+      const rebuild = async () => {
+        const entity = entityRef.current
+        if (!entity) return
+
+        try {
+          const oldComponent = componentRef.current
+          if (oldComponent) {
+            await entity.removeComponent(oldComponent)
+            await oldComponent.destroy()
+            componentRef.current = null
```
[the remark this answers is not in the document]
parthpatelsj: I have added this as a todo, and will take this as a follow up PR.

packages/react/src/reality/components/ModelEntity.tsx
```diff
@@ -1,33 +1,90 @@
...
+
+      const apply = async () => {
+        try {
+          const materialList: SpatialMaterial[] = (
+            await Promise.all(
+              next.map(mid => ctx.resourceRegistry.get<SpatialMaterial>(mid)),
+            )
+          ).filter(Boolean)
+          if (entityRef.current) {
+            await entityRef.current.setMaterials(materialList)
```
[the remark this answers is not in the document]
parthpatelsj: I have added this as a todo, and will take this as a follow up PR.

packages/react/src/reality/components/ModelEntity.tsx
parthpatelsj: I have added this as a todo, and will take this as a follow up PR.

PR #1060, opened 2026-03-24
Ua update

JackHFK:
- UA change: The VisionOS WebView customUserAgent now explicitly includes two version segments: WSAppShell/<shellVersion> and WebSpatial/<sdkVersion> . These are injected via the WS_SHELL_VERSION and WS_SDK_VERSION placeholders, preventing the legacy PACKAGE_VERSION placeholder from leaking into runtime and breaking version checks.
- CLI change: When generating/updating packages/visionOS/web-spatial/manifest.swift , the CLI replaces WS_SHELL_VERSION with the app (@webspatial/visionos ) version and WS_SDK_VERSION with the SDK version (from @webspatial/core-sdk ). The resulting values are embedded into the output so the UA composition and JS-side injection can read consistent version strings.
- Added a test for batch creation of SpatialDiv


.changeset/fluffy-peas-cheer.md
```diff
@@ -0,0 +1,7 @@
+---
+'@webspatial/platform-visionos': patch
```
khufu-dev, line 2: @JackHFK this should not be a patch update since you're updating the user agent exposed to the developer. Please update to be at least minor or major version change according to https://semver.org

Issue #1055, opened 2026-03-20
SpatialDiv popups do not receive eye/hand (gaze/hand) interaction events


parthpatelsj:
## Summary
Inside a spatialized div (SpatialDiv), popup UI triggered from within does not receive eye/hand interaction events (gaze/hand input), making it impossible to interact in XR.

## Expected
Popup UI should receive gaze/hand input like normal spatial UI.

## Actual
No gaze/hand events are delivered for the popup.

## Context
Source doc: https://bytedance.larkoffice.com/wiki/MnLYwYXhjibHrsk2Mc0cigcRnyg

PR #1090, opened 2026-03-31
expose spatial window document.body

parthpatelsj:
Fix for: https://github.com/webspatial/webspatial-sdk/issues/1055

Screenshot 2026-03-31 at 1 49 44 PM

When using Radix UI Popover (shadcn/ui) inside an enable-xr div, the popover renders to the main page's document.body instead of the spatial window's document.body. This causes the popover to appear mispositioned because:

- The trigger element lives in the spatial window's DOM (windowProxy.document.body)
- Radix portals popover content to document.body (main page)
- Position calculations reference wrong coordinate space

Approach: React Context + useSpatialPortalContainer() Hook
Expose the spatial window's document.body via React context so users can pass it to Radix's <Popover.Portal container={...}> prop.

```

import { useSpatialPortalContainer } from '@webspatial/react-sdk'
import * as Popover from '@radix-ui/react-popover'

function MyComponent() {
  const container = useSpatialPortalContainer()

  return (
    <div enable-xr>
      <Popover.Root>
        <Popover.Trigger>Open</Popover.Trigger>
        <Popover.Portal container={container}>
          <Popover.Content>Popover content here</Popover.Content>
        </Popover.Portal>
      </Popover.Root>
    </div>
  )
}
```

### DIFF
--- a/packages/react/src/spatialized-container/context/SpatialWindowContext.ts
+++ b/packages/react/src/spatialized-container/context/SpatialWindowContext.ts
@@ -1,0 +1,8 @@
+import { createContext, useContext } from 'react'
+
+export const SpatialWindowContext = createContext<WindowProxy | null>(null)
+
+export function useSpatialPortalContainer(): HTMLElement | undefined {
+  const windowProxy = useContext(SpatialWindowContext)
+  return windowProxy?.document?.body ?? undefined
+}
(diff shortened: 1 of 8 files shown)


packages/react/src/spatialized-container/context/SpatialWindowContext.ts
```diff
@@ -0,0 +1,8 @@
+import { createContext, useContext } from 'react'
+
+export const SpatialWindowContext = createContext<WindowProxy | null>(null)
+
+export function useSpatialPortalContainer(): HTMLElement | undefined {
+  const windowProxy = useContext(SpatialWindowContext)
+  return windowProxy?.document?.body ?? undefined
```
khufu-dev, line 7: Why is it falling back to undefined?
```suggestion
  return useContext(SpatialWindowContext)?.document?.body
```

khufu-dev (approved):
Very nice solution. This library luckily exposes a `Popover.Portal` component to modify the `document.body`. How would the developer know to use this component with WebSpatial, do we put this into our docs or teach our AI. 
Also what happens when a component similar to `Popover.Portal` is not exposed in other libraries?

walton007 (approved):

khufu-dev:
Closing stale pull request

PR #1101, opened 2026-04-07
openspec: runtime-feature-detection  

walton007:
## Motivation
Align the JS SDK with an explicit runtime capability contract for `WSAppShell/<semver>` (visionOS) and `PicoWebApp/<semver>` (Pico), with `supports(name, subTokens?)` resolved by a versioned capability matrix and OpenSpec runtime-feature-detection artifacts.
## Scope
**OpenSpec** — `openspec/changes/runtime-feature-detection/`: review/spec/tasks/matrix updates, including Reality unsupported fallback clarification.
**Core** — `packages/core/src/runtime/`: runtime snapshot + UA parsing, capability matrix data, `supports()` behavior/tests, DOM depth key partition, contract tests, JSB platform resolver, and `CapabilityKey` export.
Also includes `supports()` typing overload:
- typed overload for `CapabilityKey`
- string overload kept for backward-compatible unknown-key probes
**React** — runtime API exposure and typing:
- `WebSpatialRuntime.supports`
- `CapabilityKey` re-export from `@webspatial/react-sdk`
- related runtime/react coverage alignment (`useDomProxy` coverage line)
**Test server** — runtime capabilities page and contract checks UI (`apps/test-server/src/pages/runtime-capabilities/`).
**visionOS package** — runtime matrix-aligned package/changelog updates and manifest tweak.
## Out of scope (moved to follow-up PR)
To reduce review complexity, this PR does **not** include fallback gating behavior changes for:
- `useMetrics` unsupported throw behavior
- `convertCoordinate` unsupported throw behavior
- `Model` unsupported fallback gate
- model sub-capability gating in `SpatializedStatic3DElementContainer` ref extras
These are intentionally deferred to a dedicated follow-up PR.
## How to test
```bash
pnpm --filter @webspatial/core-sdk test
pnpm --filter @webspatial/react-sdk exec tsc -p tsconfig.json --noEmit
pnpm --filter @webspatial/react-sdk exec vitest run src/spatialized-container/hooks/useDomProxy.coverage.test.ts
npm run dev
```

Open **Runtime capabilities** → **Contract** / **vision vs pico** / **Live**.

## Reviewer checklist

- [ ] `CAPABILITY_TABLE` matches product matrix (1.5.0 / 1.6.0, 0.1.1 / 0.1.2).
- [ ] `supports()` + `getRuntime()` match `review.md` §3–§4.
- [ ] DOM depth: window vs ref consistent (core, React proxy, `contract-review.test.ts`, Contract UI).
- [ ] This PR does **not** remove `packages/androidXR` (separate PR).

## Screenshots

_(Attach runtime-capabilities / Contract tab screenshots here.)_
image 
image


openspec/changes/runtime-feature-detection/review.md
```diff
@@ -0,0 +1,275 @@
+# Runtime feature detection — design review (consolidated)
+
+This document consolidates agreed technical decisions and fallback conventions for **WebSpatial runtime capability detection**.
+
+<a id="review-contents"></a>
+
+## Contents
```
khufu-dev, line 7: You can omit the table of contents. Github will auto generate this.

openspec/changes/runtime-feature-detection/review.md
```diff
@@ -0,0 +1,275 @@
...
+| Field | Type | Notes |
+|---|---|---|
+| `type` | `'visionos' \| 'picoos' \| null` | `null` means not running in WebSpatial runtime. |
+| `shellVersion` | `string \| null` | Parsed from UA tokens (see §4.5). |
+
+<a id="review-2-3"></a>
+
+### 2.3 SSR / no `window`
+
+Capability checks must not throw solely because `window` is unavailable; use conservative `false`.
```
khufu-dev, line 78: What happens in SSR environment? If I use <Model> component will it wait to run on the browser to detect runtime features.

openspec/changes/runtime-feature-detection/review.md
```diff
@@ -0,0 +1,275 @@
...
+
+<a id="review-2"></a>
+
+## 2. Public API surface
+
+<a id="review-2-1"></a>
+
+### 2.1 `WebSpatialRuntime.supports(name, tokens?)`
+
+- `name` must be a documented top-level key.
```
khufu-dev, line 55: What are the types here? I'm guessing `name` is string?

openspec/changes/runtime-feature-detection/review.md
```diff
@@ -0,0 +1,275 @@
...
+- Optional `tokens` must be documented for that `name`.
```
khufu-dev, line 56: In the first version can we start without `token`. I want to understand how customers use this feature before adding too many granular features. For example if the Model sub token `autoplay` returns false would the developer really write different code or would they just ignore the feature which the Browser already does?

openspec/changes/runtime-feature-detection/review.md
```diff
@@ -0,0 +1,275 @@
...
+- true: `ready`, `currentSrc`, `entityTransform`, `paused`, `duration`, `playbackRate`, `play`, `pause`
+- false: `currentTime`
+
+If parent `supports('Model')` is `false`, all model sub-token checks must be `false`.
+
+---
+
+<a id="review-4"></a>
+
+## 4. Resolution rules
```
khufu-dev, line 206: This section is difficult to read. Can you write it with less headings and a single paragraph. Maybe it can be written succinctly as a table.

openspec/changes/runtime-feature-detection/review.md
```diff
@@ -0,0 +1,275 @@
...
+
+<a id="review-6"></a>
+
+## 6. Capability table maintenance
+
+<a id="review-6-1"></a>
+
+### 6.1 Where matrix lives
+
+- Collaboration template: [`capability-matrix.template.md`](./capability-matrix.template.md)
```
khufu-dev, line 285: Can you link to each feature document browser compatibility matrix. For example link to [Model browser compatibility](https://github.com/webspatial/webspatial-sdk/blob/main/docs/Model.md#browser-compatibility). This is similar to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/video). We can still have one global capability matrix that links out to each document.

openspec/changes/runtime-feature-detection/review.md
walton007: name is 'picoos' | 'visionos' | null

openspec/changes/runtime-feature-detection/review.md
walton007: the developer can decide what to do, we just provide those detection capability

openspec/changes/runtime-feature-detection/review.md
walton007: I didn't write this document manually. It's a AI generated md file, and I just discuss the detail with AI. You can use markdown preview to read the table and it's quite well

openspec/changes/runtime-feature-detection/review.md
walton007: Yes, we can provide a dedicated capability table generated from our internal **webspatial-capability-matrix**.
This file 'capability-matrix.template.md' is a template.

openspec/changes/runtime-feature-detection/review.md
walton007: Thanks for the question — here’s how it behaves:

Core (supports() / getRuntime()):
There’s no window/navigator on the server, so we don’t throw. getRuntime() treats that as “no WebSpatial runtime” (type: null), and supports() returns false conservatively (per §2.3). This is synchronous — there’s no async “wait for the browser.” Any call during SSR returns that conservative result immediately.

React components:
Components that matter for hydration (e.g. Model, SpatializedContainer) are wrapped with withSSRSupported. During ssr and hydrate they render a lightweight placeholder (<div> with className/style) so we don’t run the full spatial branch or rely on UA there. After hydration (after-hydrate), the real component runs in the browser, where WebSpatialRuntime.supports(...) can see navigator and return accurate results.

So: the API doesn’t “wait” — but the wrapped components defer the real detection path until client after-hydrate. If you call supports() yourself during SSR, you’ll get false until the same code runs again on the client (e.g. in an effect or after client-only render).

openspec/changes/runtime-feature-detection/review.md
walton007: We’ll keep the global capability matrix (and the spreadsheet/CSV workflow) as the single place product and engineering align on shell version × supports() key. On top of that, we can add per-feature docs (e.g. Model, WindowScene) that deep-dive behavior, edge cases, and fallbacks, and link back to the relevant row/column in the matrix so we don’t duplicate semver ranges in three places.

A small naming note: for WebSpatial, compatibility is really runtime / shell (visionOS vs Pico, WSAppShell / PicoWebApp, etc.), not Chrome-vs-Safari-style browser compatibility — we’ll phrase it that way in the docs to avoid confusion with MDN’s BCD.

Concrete URLs can point to in-repo markdown first, then to the public docs site once those pages exist. We can start with a few high-traffic APIs and expand.

packages/react/src/useMetrics.tsx
```diff
@@ -1,8 +1,15 @@
...
+import {
+  PhysicalMetrics,
+  supports,
+  WebSpatialRuntimeError,
+} from '@webspatial/core-sdk'
 
 export function useMetrics() {
   useSyncExternalStore(PhysicalMetrics.subscribe, PhysicalMetrics.getValue)
   const { pointToPhysical, physicalToPoint } = PhysicalMetrics
+  if (!supports('useMetrics')) {
```
mzvast: （普通浏览器 / SSR）：现有调用 useMetrics() 的页面/组件在非 WebSpatial runtime 下直接 crash？

packages/core/src/runtime/userAgent.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (pico?.[1]) return { version: pico[1], source: 'picoapp' }
+  return { version: null, source: null }
+}
+
+/** Pico runtime: require explicit Pico tokens (do not treat generic `like Quest` VR UAs as picoos). */
+function inferPicoOs(ua: string): boolean {
+  return (
+    /\bPicoWebApp\//i.test(ua) ||
+    /\bPicoBrowser\b/i.test(ua) ||
+    /\bPicoWebApp\b/i.test(ua)
```
JackHFK: I think there is some redundancy here, and it seems that only one of the first and third judgments is needed.

packages/core/src/runtime/jsbAdapterPlatform.ts
```diff
@@ -0,0 +1,51 @@
...
+    const diff = a[index] - b[index]
+    if (diff > 0) {
+      return true
+    }
+    if (diff < 0) {
+      return false
+    }
+  }
+  return false
+}
```
JackHFK, line 25: Perhaps we can directly use the `compareSemver` method in `semver.ts` to replace the current method.

walton007: I want to reduce third party library dependency and semver compare is quite easy for AI

packages/react/src/useMetrics.tsx
walton007: 产品定义的逻辑就是这样，需要开发者自己兜底吗，要么做特性检测，要么try/catch

packages/core/src/runtime/userAgent.ts
walton007: Yes, will fix it

packages/core/src/runtime/jsbAdapterPlatform.ts
khufu-dev: Since you're comparing a string from the browser user agent string it's better to use the semver library to handle all the edge cases. For example the  you can have a version such as "2.1" and it won't be handled properly by this custom function.

packages/react/src/webSpatialRuntime.ts
```diff
@@ -0,0 +1,9 @@
+import { supports } from '@webspatial/core-sdk'
+
+/**
+ * Runtime capability checks (`supports('Model')`, `supports('WindowScene', ['defaultSize'])`, …).
+ * Spec: `openspec/changes/runtime-feature-detection/review.md`.
+ */
+export const WebSpatialRuntime = {
+  supports,
+} as const
```
khufu-dev, line 9: Why can't we call core directly? This seems redundant.

packages/core/src/runtime/supports.ts
```diff
@@ -0,0 +1,104 @@
...
+      break
+    }
+  }
+  return chosen
+}
+
+/**
+ * Public capability probe (`WebSpatialRuntime.supports` re-exports this from React SDK).
+ */
+export function supports(name: string, tokens?: readonly string[]): boolean {
```
khufu-dev, line 70: Can you update the name and token type so only specific values are accepted instead of all strings. Right now there's no way to know which name and tokens are accepted.

packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx
```diff
@@ -152,36 +153,54 @@ function SpatializedStatic3DElementContainerBase(
...
+        Object.defineProperty(extras, 'currentSrc', {
+          enumerable: true,
+          configurable: true,
+          get(): string {
+            const spatializedElement = (domProxy as any)
+              .__spatializedElement as SpatializedStatic3DElement | undefined
+            return spatializedElement?.currentSrc ?? ''
+          },
+        })
+      }
```
khufu-dev: This doesn't work with Typescript because now the type definition says this property is available but then at runtime it crashes because it's not defined.

packages/core/src/runtime/keys.ts
```diff
@@ -0,0 +1,126 @@
+/**
+ * Canonical `supports(name)` keys (OpenSpec `review.md` §3).
+ * Aliases are normalized before lookup (see `normalizeCapabilityName`).
+ */
+
+export const COMPONENT_KEYS = [
```
khufu-dev, line 6: How are these keys used?

openspec/changes/runtime-feature-detection/review.md
```diff
@@ -0,0 +1,294 @@
...
+| Keys | Surface |
+|------|---------|
+| `xrInnerDepth`, `xrOuterDepth` | **`Window`** (global / scene readbacks) |
+| `xrClientDepth`, `xrOffsetBack` | **Spatialized container `ref`** (`HTMLElement` readbacks via the React SDK **dom proxy** — `useDomProxy.ts`; **not** `window`) |
+
+<a id="review-3-5-1"></a>
+
+#### 3.5.1 `xrInnerDepth` / `xrOuterDepth` (`Window`)
+
+When **`supports('<key>')` is `false`**, the host must **not** expose that key on **`window`’s surface** — **`!(<key> in window)`** must hold (stricter than only `window.<key> === undefined`, which still passes if an own property exists with value `undefined`).
```
khufu-dev, line 161: How does this work with the type definitions we provide? In normal web development user defines the ES version and minimum browser version and the Typescript provides the type definitions accordingly.

khufu-dev (changes requested):
I think this PR needs to be split into multiple smaller code reviews. We should start with providing the user with the foundational runtime tools such as `supports` and documenting that clearly.
Then there should be separate PRs for breaking changes to the individual components. For example `convertCoordinate` with this PR throws an exception and would be a major breaking change because the user's code would currently not handle that. 
I think as a web developer it would be easier to get a Pico OS version and then use that to condition my code. It's very difficult for developers to feature test every property on every feature and write conditional code. For example if `currentSrc` doesn't exist on Model what should developer do?

parthpatelsj:
> I think this PR needs to be split into multiple smaller code reviews. We should start with providing the user with the foundational runtime tools such as `supports` and documenting that clearly. Then there should be separate PRs for breaking changes to the individual components. For example `convertCoordinate` with this PR throws an exception and would be a major breaking change because the user's code would currently not handle that. I think as a web developer it would be easier to get a Pico OS version and then use that to condition my code. It's very difficult for developers to feature test every property on every feature and write conditional code. For example if `currentSrc` doesn't exist on Model what should developer do?

+1 - I don't think developers will write conditional code for every single component and property within component. They are more likely to developer based on checking something like https://caniuse.com/, and if browser is too old, they don't allow website to appear on the browser.

parthpatelsj:
> > I think this PR needs to be split into multiple smaller code reviews. We should start with providing the user with the foundational runtime tools such as `supports` and documenting that clearly. Then there should be separate PRs for breaking changes to the individual components. For example `convertCoordinate` with this PR throws an exception and would be a major breaking change because the user's code would currently not handle that. I think as a web developer it would be easier to get a Pico OS version and then use that to condition my code. It's very difficult for developers to feature test every property on every feature and write conditional code. For example if `currentSrc` doesn't exist on Model what should developer do?
> 
> +1 - I don't think developers will write conditional code for every single component and property within component. They are more likely to developer based on checking something like https://caniuse.com/, and if browser is too old, they don't allow website to appear on the browser.

I think this is very useful groundwork for our framework having a compatibility contract per category:

1. Self-contained spatialized additions (<Model>, <Reality>, future onrment, etc.): self-gate, render fallback or nothing. 

2. **Dynamic 3D API**: Scene graph primitives inside <Reality> (entities, materials, assets, attachments — anything referenced by id/parent): warn loudly in dev, render nothing in prod. supports() legitimately useful here for asset branching.

3. Leaf properties (CSS props, Model sub-tokens, scene options, events, future hover-effect props): silent drop, CSS-style. Native default takes over.

4. Runtime-gated JS APIs (useMetrics, convertCoordinate, session.createX): throw typed WebSpatialRuntimeError. Prefer structural absence ('x' in session returns false) when practical.

Plus: something like caniuse.com  as the dev-time reference.
supports() can stay as primary API where it matters, and as escape hatch everywhere else.

@khufu-dev, @walton007

packages/react/src/webSpatialRuntime.ts
walton007: We won't expose core sdk directly to end user

packages/core/src/runtime/jsbAdapterPlatform.ts
walton007: OK. I'll switch to third party library

packages/core/src/runtime/supports.ts
walton007: OK. I will make this change

packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx
walton007: can we change typescript definition to optional currentSrc?

packages/core/src/runtime/keys.ts
walton007: It will not expose to developers. Developers will only use sdk from webspatial react sdk. 
This is used by testserver for internal test

openspec/changes/runtime-feature-detection/review.md
walton007: is the optional xrInnerDepth work?

packages/react/src/spatialized-container/hooks/useDomProxy.ts
```diff
@@ -308,6 +314,23 @@ export class SpatialContainerRefProxy<T extends SpatializedElementRef> {
           }
           return ok
         },
+        has(target, prop) {
```
khufu-dev, line 317: I think this fixes https://github.com/webspatial/webspatial-sdk/issues/1140

khufu-dev (approved):
Thank you for making the changes. Sorry for the delayed review.

PR #1115, opened 2026-04-08
Delay stagemode orbit interaction to June

khufu-dev:
Delay stagemode orbit interaction to June since it's a lower priority than lazy loading.


khufu-dev:
> I agree, however, should we pull lazy loading to 2.0? which release on 5/5?

The lazy loading is technically complex and would be difficult to pull into 5/5 release.

PR #1125, opened 2026-04-11
ci: require changeset when PR touches packages/

walton007:
## Summary
- New workflow `validate-changeset.yml` runs on pull requests targeting `main`.
- If the PR diff touches `packages/`, it must also include a new or updated `.changeset/*.md` (from `pnpm changeset`). `.changeset/README.md` does not count.
- Maintainers can apply the `skip-changeset` label to bypass when appropriate.
- **CONTRIBUTING.md**: documents when to add a changeset, the `pnpm changeset` flow, and the bypass label (create the label once in the repo if missing).

## Implementation
- `tools/scripts/validate-pr-changeset.sh` — portable bash (no `mapfile`), compares base..head SHAs from the PR event.
- Checkout uses `head.repo.full_name` + `head.sha` so fork PRs work.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/validate-changeset.yml
+++ b/.github/workflows/validate-changeset.yml
@@ -1,0 +1,49 @@
+name: Validate changeset
+
+# When a PR changes published packages under packages/, require a .changeset/*.md
+# (add with `pnpm changeset`). Label `skip-changeset` bypasses for docs-only or maintainer exceptions.
+
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  contents: read
+  pull-requests: read
+
+concurrency:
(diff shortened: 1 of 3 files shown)


khufu-dev (approved):
Can we disable the changeset bot comments since we now have this feature? @mzvast

mzvast:
> Can we disable the changeset bot comments since we now have this feature? @mzvast

Can you explain more of what should happen with this feature?

I don't think the changeset bot comments have anything to do with this workflow.

khufu-dev:
> > Can we disable the changeset bot comments since we now have this feature? @mzvast
> 
> Can you explain more of what should happen with this feature?
> 
> I don't think the changeset bot comments have anything to do with this workflow.

I thought this workflow would fail if there was no new changeset which is is similar to the changeset bot comment.

PR #1129, opened 2026-04-13
End-to-end texture support for UnlitMaterial

parthpatelsj:
The SDK had scaffolding for textures but none of it was wired up — textureId on materials was dead code. This PR makes textures a real, loadable resource across all three layers.

A new <Texture id url> React component handles the lifecycle: it resolves URLs, creates the texture via the core SDK, and registers a promise in ResourceRegistry. UnlitMaterial now resolves textureId through the registry to get the native UUID before sending it over the JSB bridge 

On the Swift side, SpatialTextureResource now actually downloads the image (reusing loadResourceToLocal) and decodes it with TextureResource(contentsOf:) on MainActor. Creation is async — the JS promise only resolves once the texture is fully loaded and usable. Runtime URL changes re-download and cascade refreshMaterials() across all scene objects.

Also includes a test page (/reality/textured-unlit-box), PRD updates adding a Textures section, and the CreateTextureResource → CreateTexture command rename that would have silently failed on the bridge.
Screenshot 2026-04-27 at 1 31 55 PM


walton007:
@codex review

walton007:
@codex please review

packages/visionOS/web-spatial/model/SpatialScene.swift
```diff
@@ -954,8 +956,27 @@ class SpatialScene: SpatialObject, ScrollAbleSpatialElementContainer, WebMsgSend
         //      @fukang: add Component here
     }
 
+    private func onCreateTexture(command: CreateTexture, resolve: @escaping JSBManager.ResolveHandler<Encodable>) {
+        let texture = SpatialTextureResource(command.url)
+        addSpatialObject(texture)
+        Task {
```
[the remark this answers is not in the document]
parthpatelsj: We register the texture before load() for the same command flow, but on load failure we call texture.destroy(), so the SpatialTextureResource is removed from the scene and doesn’t accumulate. 

JS only gets an id on success either way.

packages/react/src/reality/utils/ResourceRegistry.ts
```diff
@@ -1,44 +1,79 @@
...
+    if (existing) {
+      return existing as Promise<T>
+    }
+
+    // Nothing registered yet — return a promise that add() will resolve (or reject on failure).
+    const promise = new Promise<SpatialObject>((resolve, reject) => {
+      this.deferreds.set(id, { resolve, reject })
+    })
+    this.resources.set(id, promise)
+    return promise as Promise<T>
```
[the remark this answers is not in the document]
parthpatelsj: No other code uses a timed rejection for deferred lookups. This is a great suggestion, but I would move this to another PR.

packages/react/src/reality/components/Texture.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }
+    init()
+
+    return () => {
+      controller.abort()
+      resourceRegistry.removeAndDestroy(options.id)
+      textureRef.current = undefined
+      isInitializedRef.current = false
+    }
+  }, [ctx])
```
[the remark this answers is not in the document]
parthpatelsj: This is now addressed. 
1) I've put `options.url` on the same effect that creates the texture: [ctx, options.id, options.url].
2) When url changes, React runs cleanup first (abort + removeAndDestroy), then starts a new create for the new URL. The in-flight create from the old URL should see abort and destroy on completion.

docs/dynamic-3d-api-prd.md
```diff
@@ -8,10 +8,10 @@
 | Approach | Component / API | Use case |
 |----------|-----------------|----------|
 | **Reality** | `<Reality>` + scene graph | 3D canvas: shapes (primitives) and model instances; like `<canvas>` but for 3D |
-| **Model (static)** | `<Model enable-xr src="…">` | Drop-in 3D model file; like `<img>` for 3D |
+| **Model (static)** | `<Model enable-xr src="…">` | Drop-in 3D model file; like `<img>` for 3D — see **`docs/Model.md`** (not covered here) |
```
khufu-dev, line 11: Let's remove this section to avoid confusion. We'll add a table of contents at the root of docs

docs/dynamic-3d-api-prd.md
```diff
@@ -251,86 +284,9 @@ useEffect(() => {
...
-| `RemoveComponentFromEntity` | JS → Native | Remove component from entity (used before geometry rebuild) |
-| `SetMaterialsOnEntity` | JS → Native | Override materials on a model entity |
-
----
-
-## 19. References
+## 14. References
 
-- **Model (static):** `docs/Model.md`
+- **Standalone `<Model>` (not part of this spec):** `docs/Model.md`
```
khufu-dev, line 289: Remove

packages/react/src/reality/components/Texture.tsx
```diff
@@ -0,0 +1,59 @@
...
+    url.startsWith('https://') ||
+    url.startsWith('file://')
+  ) {
+    return url
+  }
+  if (typeof window === 'undefined' || !window.location?.href) {
+    return url
+  }
+  return new URL(url, window.location.href).href
+}
```
khufu-dev: Aren't we converting to absoluteURL in multiple places in the codebase? We should probably create a urlUtils file and put it there.

packages/react/src/reality/components/Texture.tsx
```diff
@@ -0,0 +1,59 @@
...
+  url: string
+  onLoad?: () => void
+  onError?: (error: any) => void
+}
+
+function resolveTextureUrl(url: string): string {
+  if (
+    url.startsWith('http://') ||
+    url.startsWith('https://') ||
+    url.startsWith('file://')
```
khufu-dev: Why would it start with `file://` that doesn't make sense to me?

packages/react/src/reality/components/Texture.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export type TextureProps = {
+  children?: React.ReactNode
+  id: string
+  url: string
+  onLoad?: () => void
+  onError?: (error: any) => void
+}
+
+function resolveTextureUrl(url: string): string {
```
khufu-dev:
```suggestion
function getAbsoluteURL(url: string): string {
```

packages/react/src/reality/components/Texture.tsx
```diff
@@ -0,0 +1,59 @@
...
+        const resolvedUrl = resolveTextureUrl(options.url)
+        const texturePromise = session.createTexture({ url: resolvedUrl })
+        resourceRegistry.add(textureId, texturePromise)
+        const texture = await texturePromise
+        if (controller.signal.aborted) {
+          texture.destroy()
+          return
+        }
+        options.onLoad?.()
+      } catch (error: any) {
```
khufu-dev:
```suggestion
      } catch (error: unknown) {
```

packages/react/src/reality/components/Texture.tsx
```diff
@@ -0,0 +1,59 @@
...
+        options.onError?.(error)
+      }
+    }
+    init()
+
+    return () => {
+      controller.abort()
+      resourceRegistry.removeAndDestroy(textureId)
+    }
+  }, [ctx, options.id, options.url])
```
khufu-dev:
```suggestion
  }, [ctx, id, url])
```

packages/react/src/reality/components/Texture.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }
+  if (typeof window === 'undefined' || !window.location?.href) {
+    return url
+  }
+  return new URL(url, window.location.href).href
+}
+
+export const Texture: React.FC<TextureProps> = ({ children, ...options }) => {
+  const ctx = useRealityContext()
+
```
khufu-dev, line 21:
```suggestion
const {id, url} = options
```

packages/react/src/reality/components/UnlitMaterial.tsx
```diff
@@ -19,11 +20,32 @@ export const UnlitMaterial: React.FC<UnlitMaterialProps> = ({
...
     if (!ctx) return
-    const { session, reality, resourceRegistry } = ctx
+    const { session, resourceRegistry } = ctx
     const init = async () => {
-      const materialPromise = session.createUnlitMaterial(options)
-      resourceRegistry.add(options.id, materialPromise)
       try {
+        let textureIdForNative = options.textureId
+        if (options.textureId) {
+          const texturePending = resourceRegistry.get<SpatialObject>(
```
khufu-dev:
```suggestion
          const texturePromise = resourceRegistry.get<SpatialObject>(
```

packages/react/src/reality/components/UnlitMaterial.tsx
```diff
@@ -19,11 +20,32 @@ export const UnlitMaterial: React.FC<UnlitMaterialProps> = ({
...
+            options.textureId,
+          )
+          if (texturePending) {
+            try {
+              const textureResource = await texturePending
+              textureIdForNative = textureResource.id
+            } catch (error) {
+              console.error(' ~ UnlitMaterial ~ texture resolve error:', error)
```
khufu-dev: What type of error is expected here? Is it a loading error. I want to make sure that we don't pollute the error log.

packages/react/src/reality/components/UnlitMaterial.tsx
```diff
@@ -43,16 +65,55 @@ export const UnlitMaterial: React.FC<UnlitMaterialProps> = ({
...
+            } else {
+              updates.textureId = options.textureId
+            }
+          }
+        }
+        if (options.transparent !== undefined)
+          updates.transparent = options.transparent
+        if (options.opacity !== undefined) updates.opacity = options.opacity
+        if (cancelled || Object.keys(updates).length === 0) return
+        await materialRef.current?.updateProperties(updates)
```
khufu-dev: I don't think the await is necessary?

packages/react/src/reality/utils/ResourceRegistry.ts
```diff
@@ -1,44 +1,79 @@
...
+        )
+      this.deferreds.delete(id)
+    }
   }
-  remove(id: string) {
+
+  get<T extends SpatialObject>(id: string): Promise<T> {
+    const existing = this.resources.get(id)
+    if (existing) {
+      return existing as Promise<T>
```
khufu-dev: This cast could be dangerous

packages/react/src/reality/utils/ResourceRegistry.ts
khufu-dev, line 1: Do we really have to keep track of promises instead of the actual resources? Promises become very complicated when used with React.

packages/react/src/reality/utils/ResourceRegistry.ts
```diff
@@ -1,44 +1,79 @@
...
-
-    // Best-effort destroy for all resolved and future-resolving resources
-    pending.forEach(promise =>
-      promise
-        .then(spatialObj => spatialObj.destroy())
-        .catch(() => {
-          // swallow rejection to avoid unhandled promise errors during teardown
-        }),
-    )
+    pending.forEach(p => p.then(obj => obj.destroy()).catch(() => {}))
```
khufu-dev: This becomes very messy since the Promises could be cleaning up much later after the component is unmounted.

packages/visionOS/web-spatial/model/SpatialScene.swift
```diff
@@ -1233,20 +1255,71 @@ class SpatialScene: SpatialObject, ScrollAbleSpatialElementContainer, WebMsgSend
...
+                    refreshedMaterialIds.insert(material.spatialId)
+                }
+                for mid in refreshedMaterialIds {
+                    refreshComponentsUsingMaterial(mid)
+                }
+                resolve(.success(baseReplyData))
+            } catch {
+                resolve(.failure(JsbError(code: .CommandError, message: error.localizedDescription)))
+            }
+        }
```
khufu-dev, line 1282: SpatialScene should not have a async code like this. I think maybe the architecture of this is not quite correct. In this case `onUpdateTextureProperties` is directly waiting for all the code to finish before returning a result. Instead you may want to do one command to start the process and then a callback command to return the results. In any case this code should at least go into another file.

packages/visionOS/web-spatial/model/SpatialScene.swift
```diff
@@ -1233,20 +1255,71 @@ class SpatialScene: SpatialObject, ScrollAbleSpatialElementContainer, WebMsgSend
         resolve(.success(baseReplyData))
     }
 
+    /// After `SpatialUnlitMaterial.updateProperties`, ModelComponent still has a stale material copy — only refresh users of that material id.
+    private func refreshComponentsUsingMaterial(_ materialId: String) {
```
khufu-dev: This probably should go into a different file not SpatialScene maybe SpatialMaterial

packages/visionOS/web-spatial/model/dynamic3d/SpatialTextureResource.swift
```diff
@@ -3,15 +3,40 @@ import SwiftUI
...
         _resource
     }
 
     override init(_ url: String) {
+        self.url = url
         super.init()
     }
 
+    func load() async throws {
+        let localURL = try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<URL, Error>) in
```
khufu-dev, line 18: When the user navigates to another page are these asynchronous loads automatically cancelled and cleaned up?

khufu-dev (approved):
It was a big change and I tried to review as much as possible.

packages/react/src/reality/components/Texture.tsx
parthpatelsj: Good point file:// is not needed.

packages/react/src/reality/components/Texture.tsx
parthpatelsj: Added getAbsoluteUrl in utils/urlUtils.ts and wired Texture and ModelAsset to it (they were the two with the same “absolute for the bridge” pattern).

packages/react/src/reality/components/UnlitMaterial.tsx
parthpatelsj: This throws if `CreateTexture` failed on native side, `ResourceRegistry` was torn down, or cancelled work. You make a good point about polluting error log, and `<Texture>`  should surface the load errors `onError`. So this is redundant/noisy.

packages/react/src/reality/utils/ResourceRegistry.ts
parthpatelsj: it should return `Promise <SpatialObject>`

packages/react/src/reality/utils/ResourceRegistry.ts
parthpatelsj: Resources are created asynchronously (createTexture, createMaterial, etc.), so the registry has to represent "in flight" somehow. 

A Promise<SpatialObject> is the minimal representation: it lets a sibling call get(id) before the producer's add(id, ...) resolves, without a separate layer

The deferred branch exists specifically to keep JSX sibling order independent — 
`<Material textureId="x">` mounting before 
`<Texture id="x">` still works. 

Other options are either synchronous creation (not possible through JSB) or a parallel pub/sub mechanism for "wait until ready" or something else...

packages/react/src/reality/utils/ResourceRegistry.ts
parthpatelsj: This is hard to avoid - without aborting in-flight native work. I think this brings up a good point - if we want no late destroys, we need to explore another solution

packages/visionOS/web-spatial/model/SpatialScene.swift
parthpatelsj: 1) On file location — agreed, SpatialScene shouldn't own texture update logic. I'll move refreshComponentsUsingMaterial and the update orchestration into SpatialMaterial (or a dedicated updater).

2) On start + callback vs await — I'd push back here. The JSB convention for resource commands is 
await-in-handler with bounded async work: 
`CreateModelAsset` and `CreateSpatializedStatic3DElement` (with modelURL) both download + decode inside the handler before resolving. 

Start + callback is reserved for genuinely long-lived contexts like for <SpatialDiv> & <Attachment>— creating a new window with its own lifecycle.

Texture download + decode is bounded work, no progress events needed, and JS just wants await updateTextureProperties() to mean "it's done." 

Splitting into start + callback adds the same work as spatial div and attachments for the same observable behavior.

If this is an issue, maybe we can address this in follow up PR

packages/visionOS/web-spatial/model/dynamic3d/SpatialTextureResource.swift
parthpatelsj: Navigation destroys spatial objects and clears the texture’s TextureResource reference, but in-flight loadResourceToLocal / URLSession work is not tied to object lifetime, so it is not automatically cancelled

PR #1136, opened 2026-04-14
fix: `SpatialWebEvent` should supports multiple receivers per id

mzvast:
## Summary

Fixes an issue where multiple React components using `useMetrics()` could break each other’s subscriptions to physical metrics updates.

Symptoms:

- The most recently mounted component would “win” and earlier subscribers stopped receiving updates.
- Unmounting any one subscriber could remove the shared listener and cause remaining subscribers to stop receiving updates.

## Root Cause

`PhysicalMetrics.subscribe()` registers to `SpatialWebEvent` using a fixed id (`"window"`).

Before this change, `SpatialWebEvent` only supported a single receiver per id:

- `addEventReceiver(id, cb)` overwrote any existing receiver for that id.
- `removeEventReceiver(id)` deleted the receiver for that id entirely.

With multiple `useSyncExternalStore` subscriptions, this resulted in receiver overwrite and accidental removal.

## Changes

- `SpatialWebEvent` now supports multiple receivers per id.
- `SpatialWebEvent.removeEventReceiver(id, callback?)` now accepts an optional `callback`:
  - If `callback` is provided: remove only that receiver; delete the id entry only when it becomes empty.
  - If `callback` is omitted: keep existing behavior and remove all receivers for that id.
- `PhysicalMetrics.subscribe()` now unsubscribes by removing only its own handler from the `"window"` receivers.

## Compatibility Notes

- Existing call sites using `removeEventReceiver(id)` continue to remove all receivers for the id.
- For ids that only register a single receiver (the common case), behavior remains effectively unchanged.

## Tests

- Added coverage to verify multiple concurrent `PhysicalMetrics.subscribe()` subscribers receive updates and unsubscribing one does not affect the other.

Commands:

```bash
pnpm -C packages/core run lint
pnpm test
```

### DIFF
--- a/packages/core/src/SpatialWebEvent.ts
+++ b/packages/core/src/SpatialWebEvent.ts
@@ -4,20 +4,35 @@ interface SpatialWebEventData {
 }
 
 export class SpatialWebEvent {
-  static eventReceiver: Record<string, (data: any) => void> = {}
+  static eventReceiver: Record<string, Set<(data: any) => void>> = {}
   static init() {
     // inject __SpatialWebEvent
     window.__SpatialWebEvent = ({ id, data }: SpatialWebEventData) => {
       // console.log('__SpatialWebEvent', id, data)
-      SpatialWebEvent.eventReceiver[id]?.(data)
+      const receivers = SpatialWebEvent.eventReceiver[id]
+      if (!receivers) return
+      receivers.forEach(receiver => receiver(data))
     }
   }
 
   static addEventReceiver(id: string, callback: (data: any) => void) {
-    SpatialWebEvent.eventReceiver[id] = callback
+    // A single `id` can have multiple receivers.
+    SpatialWebEvent.eventReceiver[id] ??= new Set()
+    SpatialWebEvent.eventReceiver[id].add(callback)
   }
 
-  static removeEventReceiver(id: string) {
-    delete SpatialWebEvent.eventReceiver[id]
+  static removeEventReceiver(id: string, callback?: (data: any) => void) {
+    // If callback is omitted, remove all receivers for the id.
+    if (!callback) {
+      delete SpatialWebEvent.eventReceiver[id]
+      return
+    }
+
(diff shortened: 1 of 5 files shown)


khufu-dev:
@codex review please

packages/core/src/SpatialWebEvent.ts
```diff
@@ -4,20 +4,35 @@ interface SpatialWebEventData {
...
-  static eventReceiver: Record<string, (data: any) => void> = {}
+  static eventReceiver: Record<string, Set<(data: any) => void>> = {}
   static init() {
     // inject __SpatialWebEvent
     window.__SpatialWebEvent = ({ id, data }: SpatialWebEventData) => {
       // console.log('__SpatialWebEvent', id, data)
-      SpatialWebEvent.eventReceiver[id]?.(data)
+      const receivers = SpatialWebEvent.eventReceiver[id]
+      if (!receivers) return
+      receivers.forEach(receiver => receiver(data))
```
khufu-dev: Could this be reduced to?
```suggestion
      SpatialWebEvent.eventReceiver[id]?.forEach(fn => fn(data))
```

packages/core/src/SpatialWebEvent.ts
```diff
@@ -4,20 +4,35 @@ interface SpatialWebEventData {
 }
 
 export class SpatialWebEvent {
-  static eventReceiver: Record<string, (data: any) => void> = {}
+  static eventReceiver: Record<string, Set<(data: any) => void>> = {}
```
khufu-dev: Should this be a Map instead of Record?

khufu-dev (approved):

packages/core/src/SpatialWebEvent.ts
mzvast: Great, updated!

packages/core/src/SpatialWebEvent.ts
mzvast: Much cleaner!

PR #1137, opened 2026-04-14
feat: SpatialDiv content-ready lifecycle + demo

walton007:
## Summary

This PR includes OpenSpec updates plus implementation for the SpatialDiv content-ready lifecycle.

- Introduce SpatialDiv-only `onSpatialContentReady(ctx)` in `@webspatial/react-sdk` (with optional cleanup).
- Ensure ready callback timing follows layout-effect semantics and never fires during render.
- Guarantee ref readiness ordering and deduplicate ref dispatches to avoid repeated same-value/null emissions.
- Update fallback behavior: in non-WebSpatial fallback, `onSpatialContentReady` still fires with connected fallback host; in attachment-degraded fallback, callback remains suppressed.
- Clarify nested callback ordering contract: guaranteed parent-before-child in WebSpatial runtime; unspecified in plain web fallback.
- Add/extend tests and OpenSpec docs/tasks for acceptance matrix and runtime-dependent ordering note.
- Extend `apps/test-server/src/pages/spatial-content-ready-three/index.tsx` demo with clearer acceptance signals, safer Three.js mount slot sizing, and purple-panel click toggle (pause/resume) with event logs.
- Include changeset for `@webspatial/react-sdk` release note.

## Test plan

- `pnpm --filter @webspatial/react-sdk test -- coverage-boost.test.ts`
- `npm run test` in `apps/test-server`
- Manual verification: open `/spatial-content-ready-three` and validate:
  - nested ordering/cleanup/ref indicators
  - purple panel click toggles animation state and appends toggle logs

## Screenshots
image


parthpatelsj:
Without this change, currently children inside `<div enable-xr>` only mount once the portal WebView is ready, so useEffect in a child component already fires at the correct time. 

In that case, is this needed?
@walton007

openspec/changes/spatialdiv-content-ready-lifecycle/proposal.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Why
+
+`enable-xr` / `SpatialDiv` does not behave like a plain `div` at the DOM level: the SDK maintains a hidden layout/sampling host tree and a separate visible content tree rendered via a portal / isolated `windowProxy`. When developers follow the common web pattern of `ref` + `useEffect([])` + child DOM `ref`, they can hit timing issues (`ref` is still `null`), miss later transitions, or attach external renderers (for example a Three.js `WebGLRenderer` canvas) to a host that is not the visible content surface.
+
+We need an explicit, testable, and documentable lifecycle hook that fires when the visible content host is ready for imperative mounting, and that supports reliable teardown across unmounts, StrictMode remount probes, and internal recreation.
+
+## What Changes
+
+- Add a React prop: `onSpatialContentReady?: (ctx) => void | (() => void))` that notifies app code when the SpatialDiv **visible content host** is ready, and supports returning a cleanup function.
```
khufu-dev, line 11: I'm worried this design will be incompatible with current and future [web standards](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/load_event). Normal `div` creation is treated as a synchronous process and future [spatial CSS](https://webkit.github.io/explainers/css-spatial/explainer.html) proposal will likely keep that paradigm without a `onReady` callback.

khufu-dev (changes requested):

openspec/changes/spatialdiv-content-ready-lifecycle/proposal.md
walton007: I think the key distinction here is that this API is not trying to model plain div creation or introduce a new DOM-standard lifecycle.

The problem we are solving is specific to WebSpatial's current implementation a SpatialDiv has a hidden layout/sampling tree plus a separate visible content host rendered through the portal / windowProxy path. For imperative integrations like Three.js the important event is not "the div exists" but "the visible content host is connected and safe to mount into".

So I agree we should be careful not to present this as a generic web-standard onReady pattern. I can tighten the proposal wording to make that explicit this is a WebSpatial-specific readiness hook for the visible content host not a claim about how future spatial CSS should work.

walton007:
@parthpatelsj I think it is still needed for the cases this PR is targeting.

If app code only needs "some child effect after mount" then the current behavior can be enough. But for imperative integrations like Three.js the important contract is stricter app code needs the visible content host not just a child component mount and it also needs a reliable cleanup / re-ready signal when that host is recreated or replaced.

That is the gap this API is trying to close. The /spatial-content-ready-three demo is meant to show those cases more concretely mounting to the right host ref ordering cleanup and nested behavior.

khufu-dev (approved):

khufu-dev:
@codex review please

khufu-dev:
In the future can we split the technical documents and code into separate pull requests to make code reviews easier?

packages/react/src/spatialized-container/hooks/useSpatialContentReady.ts
```diff
@@ -0,0 +1,72 @@
...
+    spatializedElement,
+    portalInstanceObject,
+    hostElement,
+    onSpatialContentReady,
+  } = params
+
+  const callbackRef = useRef(onSpatialContentReady)
+  callbackRef.current = onSpatialContentReady
+
+  useLayoutEffect(() => {
```
parthpatelsj, line 41: I believe `useLayoutEffect` fires bottom up (children before parents). However, spec mentions 

> "In WebSpatial runtime (portal path), for nested SpatialDiv, parent onSpatialContentReady MUST run before child onSpatialContentReady on the same rising-edge transition where both become ready."

What is desired behavior?

walton007:
> In the future can we split the technical documents and code into separate pull requests to make code reviews easier?

AI wrote code some quickly when design is confirmed. I can review the code and run the demo locally to verfiy the result very quickly. 
When I iterate the requirement, AI can change both the spec and code at the same time. 
If I seperate to 2 different PRs, it's not as convenient as now  to do the iteration

parthpatelsj (approved):

PR #1154, opened 2026-04-22
test(react): cover in-operator support for extra ref props

### LINKED ISSUE #1140: in operator fails for extraRefProps properties on spatialized element refs
The Proxy created in SpatialContainerRefProxy.updateStandardSpatializedContainerDom defines get and set traps that correctly delegate to extraRefProps, but does not define a has trap. This causes the in operator to return false for properties like entityTransform, currentSrc, and ready — even though direct property access works fine.

Reproduction
```
const modelRef = useRef<ModelRef>(null);

// ❌ Always false — falls through to Reflect.has(domElement, 'entityTransform')
if (modelRef.current && "entityTransform" in modelRef.current) {
  // never reached
}

// ✅ Works — handled by the Proxy get trap
if (modelRef.current?.entityTransform !== undefined) {
  // works
}
```

[useDomProxy.ts L109–L253](https://github.com/webspatial/webspatial-sdk/blob/086ee5585ca932629eecd1dc529ecf08ba4e72f8/packages/react/src/spatialized-container/hooks/useDomProxy.ts#L109-L253)

The Proxy handler has get and set traps that check extraRefProps for additional properties (e.g. entityTransform on Model refs). However, without a has trap, the in operator falls through to Reflect.has(target, prop), which checks the underlying DOM element. The DOM element doesn't have these properties, so in always returns false.

Affects all properties injected via extraRefProps on any spatialized ref. Currently this includes Model refs (entityTransform, currentSrc, ready), but would apply to any future component that uses this pattern. The in operator is a standard JS idiom for property checking, so this is likely to be a recurring source of confusion.

walton007:
## Summary

Adds a regression test for issue #1140.

## Changes

- adds coverage for the `in` operator on ref proxy `extraRefProps`
- verifies injected ref properties are discoverable via proxy `has` semantics
- keeps production code unchanged because `main` already includes the runtime `has` trap fix

## Notes

This PR is intentionally test-only and documents the expected behavior so the issue does not regress.

Closes #1140

### DIFF
--- a/packages/react/src/spatialized-container/hooks/useDomProxy.coverage.test.ts
+++ b/packages/react/src/spatialized-container/hooks/useDomProxy.coverage.test.ts
@@ -151,6 +151,24 @@ describe('SpatialContainerRefProxy', () => {
 
     expect((ref.current as any).foo).toBe('bar')
   })
+
+  it('supports the in operator for extra ref props', () => {
+    const ref = { current: null as any }
+    const proxy = new SpatialContainerRefProxy<any>(ref, () => ({
+      get foo() {
+        return 'bar'
+      },
+    }))
+
+    const dom = document.createElement('div')
+    const task = document.createElement('div')
+
+    proxy.updateStandardSpatializedContainerDom(dom)
+    proxy.updateTransformVisibilityTaskContainerDom(task)
+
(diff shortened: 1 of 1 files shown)


khufu-dev (approved):

khufu-dev:
Closing stale PR

PR #1156, opened 2026-04-22
fix a link nested issue

mzvast:
# Scene Polyfill Anchor Navigation

The scene polyfill installs a document click handler via `hijackWindowATag(...)` to handle navigation for anchor tags in a controlled way.

## What It Does

- It listens to `document.onclick` and walks up the DOM tree from the event target to find the nearest `<a>` element.
- If the anchor has a `target` and `target !== '_self'`, it prevents the default navigation and calls `window.open(url, target)`.
- If no special handling is needed, it lets the browser behavior continue.

## Nested Clicks

Clicks often originate from a nested element inside an anchor, for example:

```html
<a href="https://example.com" target="_blank">
  <img src="..." />
</a>
```

In this case the `event.target` is the ``, not the `<a>`. The polyfill therefore uses the anchor found during event bubbling rather than relying on `event.target` being the anchor itself.

### DIFF
--- a/docs/scene-polyfill-anchor-navigation.md
+++ b/docs/scene-polyfill-anchor-navigation.md
@@ -1,0 +1,21 @@
+# Scene Polyfill Anchor Navigation
+
+The scene polyfill installs a document click handler via `hijackWindowATag(...)` to handle navigation for anchor tags in a controlled way.
+
+## What It Does
+
+- It listens to `document.onclick` and walks up the DOM tree from the event target to find the nearest `<a>` element.
+- If the anchor has a `target` and `target !== '_self'`, it prevents the default navigation and calls `window.open(url, target)`.
+- If no special handling is needed, it lets the browser behavior continue.
+
+## Nested Clicks
+
+Clicks often originate from a nested element inside an anchor, for example:
+
+```html
+<a href="https://example.com" target="_blank">
+  <img src="..." />
+</a>
+```
+
+In this case the `event.target` is the `<img>`, not the `<a>`. The polyfill therefore uses the anchor found during event bubbling rather than relying on `event.target` being the anchor itself.
--- a/packages/core/src/scene-polyfill.ts
+++ b/packages/core/src/scene-polyfill.ts
@@ -526,7 +526,7 @@ export function hijackWindowATag(openedWindow: WindowProxy) {
 
     // Look for <a> element in the clicked elements parents and if found override navigation behavior if needed
     while (element) {
-      if (element && element.tagName == 'A') {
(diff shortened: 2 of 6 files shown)


docs/scene-polyfill-anchor-navigation.md
```diff
@@ -0,0 +1,21 @@
+# Scene Polyfill Anchor Navigation
```
khufu-dev: Maybe we should call this document WebSpatial Quirks. It documents abnormal behaviour that normal web developers would expect. We can include details about anchor navigation, spatial div clipping and other abnormalities.

packages/core/src/scene-polyfill.ts
```diff
@@ -523,17 +523,16 @@ export function hijackWindowOpen(window: WindowProxy) {
 export function hijackWindowATag(openedWindow: WindowProxy) {
   openedWindow!.document.onclick = function (e) {
     let element = e.target as HTMLElement | null
-    let found = false
 
     // Look for <a> element in the clicked elements parents and if found override navigation behavior if needed
-    while (!found) {
+    while (element) {
       if (element && element.tagName == 'A') {
```
khufu-dev: while loop already checks if element is truthy.
```suggestion
      if (element.tagName == 'A') {
```

khufu-dev (approved):

khufu-dev:
@codex review

walton007:
@codex review

docs/scene-polyfill-anchor-navigation.md
walton007: Agree. We should setup a WebSpatial Quirks document

mzvast: ok，renamed

packages/core/src/scene-polyfill.ts
mzvast: cleaned the code

packages/core/src/scene-polyfill.ts
```diff
@@ -548,18 +547,15 @@ export function hijackWindowATag(openedWindow: WindowProxy) {
...
-    }
+function handleATag(event: MouseEvent, link: HTMLAnchorElement) {
+  // Use the anchor found during bubbling so nested clicks like <a><img /></a> are handled correctly.
+  const target = link.target
+  const url = link.href
+
+  if (target && target !== '_self') {
+    event.preventDefault()
+    window.open(url, target)
+    return true
```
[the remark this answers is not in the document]
mzvast: accepted and add test for it

PR #1171, opened 2026-04-28
test: add React 18 and 19 compatibility CI fixtures

walton007:
## Summary

Add explicit React compatibility coverage for `@webspatial/react-sdk` so the SDK's `jsxImportSource` integration is verified against both React 18 and React 19 host apps.

- Add `tests/react18-compat`, a Vite 6 + React 18 fixture that runs `tsc` and `vite build`.
- Add `tests/react19-compat`, a Vite 8 + React 19 fixture that runs `tsc` and `vite build`.
- Add `test:react-compat` at the root and wire it into GitHub Actions CI after the existing test step.
- Pin `react`, `react/jsx-runtime`, and `react/jsx-dev-runtime` type resolution to each fixture's host app types. This models the intended host-app-wins behavior and prevents a workspace/local SDK checkout's React 18 dev types from leaking into a React 19 app.

## Why

React 19 apps can hit `StrictMode cannot be used as a JSX component` when a linked/workspace SDK checkout exposes its own React 18 dev types. The fixtures make that compatibility contract explicit and keep React 18 support covered at the same time.

## Validation

- `pnpm run buildPackages`
- `pnpm run test:react-compat`
- `pnpm run test`

## Notes

This was built in an isolated worktree (`/tmp/webspatial-sdk-react-compat`) to avoid touching existing uncommitted changes in the main local SDK checkout.


khufu-dev (approved):
I think we should be upgrading our dependencies aggressively since we're building leading edge experimental technologies on brand new technology. 
React 19 came out in 2024 and we need to upgrade soon at which point React 18 will no longer be supported.

walton007:
Thanks, I agree we should keep moving toward React 19 and make sure the SDK works well with the latest React stack.

At the same time, React 18 still has a very large installed base. According to the State of React 2025 survey, React 18.x is still the daily React version for 1,089 out of 2,636 respondents, which is about 41.3%, while React 19.x is about 48.4%: https://2025.stateofreact.com/en-US/usage/

Given that React 18 is still used by roughly four out of ten React developers, I think we should continue supporting React 18 for now while also adding/maintaining React 19 compatibility. That gives us a migration path without cutting off a significant portion of current users.

PR #1173, opened 2026-04-28
chore: upgrade minimum ECMAScript version to 2024

khufu-dev:


.changeset/solid-garlics-exist.md
```diff
@@ -0,0 +1,10 @@
+---
+'autotest': major
+'web-content': major
+'@webspatial/react-sdk': major
```
mzvast, line 4: Do we break any public API? Or do we used any es2024 feature and exported to users?
Maybe this can be minor change for now. 
We may mark it as major when we really introduced es2024 features and exposed API to user in the future.

tests/ci-test/tsconfig.app.json
```diff
@@ -1,9 +1,9 @@
 {
   "compilerOptions": {
     "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
-    "target": "ES2020",
+    "target": "ES2024",
```
mzvast, line 4: As we are using tsup and ,according to this [issue](https://github.com/egoist/tsup/issues/1152), changes on tsconfig.target has no effect on build output, which always use `esnext`.

packages/react/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
-    "target": "ES2023",
+    "target": "ES2024",
```
mzvast, line 3: As we are using tsup and ,according to this https://github.com/egoist/tsup/issues/1152, changes on tsconfig.target has no effect on build output, which always use esnext

mzvast (commented):
What benefit can we get from es2024?

.changeset/solid-garlics-exist.md
khufu-dev: Changing the minimum version is a public API change since it only targets browsers that support ES2024 or newer. I'll start using ES2024 features soon after this change.

tests/ci-test/tsconfig.app.json
khufu-dev: Good to know. I updated it in case some IDEs may use this configuration for intellisense.

packages/react/tsconfig.json
khufu-dev: Good to know. I updated it in case some IDEs may use this configuration for intellisense.

khufu-dev:
> What benefit can we get from es2024?

Mainly to target new Javascript features and optimize the output bundle. Since we don't support old browsers we can upgrade to recent versions of ECMAScript.

.changeset/solid-garlics-exist.md
walton007: I agree this does not need to be a major bump in this PR.

At the moment the change only updates TypeScript / ESLint configuration and does not change public APIs or introduce ES2024-only runtime features into the SDK surface. Also, for the `core` / `react` packages, the actual bundle output may not be materially affected by `tsconfig.target` because they are built through `tsup`.

I would suggest treating this as a patch-level chore for now. We can reserve the major bump for the follow-up where we actually start emitting / exposing ES2024-only behavior, or where we explicitly update the SDK supported runtime baseline and document it as a breaking change in the release notes.

khufu-dev: I converted to minor version change but Codex https://github.com/webspatial/webspatial-sdk/pull/1173#discussion_r3170427636 is recommending this to be a major version bump. I'm ok with either one, long term we should not be worried about shipping major version changes.

.changeset/eleven-months-tickle.md
```diff
@@ -0,0 +1,10 @@
+---
+'autotest': minor
+'web-content': minor
+'@webspatial/react-sdk': minor
```
parthpatelsj: This should be major right? Since bumping minimum ECMAScript version is breakdown change for downstream constumers?

khufu-dev: There was a comment earlier to make it minor https://github.com/webspatial/webspatial-sdk/pull/1173#discussion_r3158342610 but I think it should be major.

parthpatelsj (approved):

.changeset/eleven-months-tickle.md
mzvast: Yeah, this case is of a little edge case. I think major and minor bump are all acceptable.

[reason for major]
If this PR changed output and includes es2024 code we can surely mark it major, and that's done, next PR should not care about es2024 any more.

[reason for not major]
It not really breaking any API and no es2024 code shipped.
The **down side** is next PR we are very likely to ship es2024 code  without realizing it should be major change.
It's hard to remember to mark major when used es2024 code.

mzvast (approved):

packages/cli/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "allowJs": true,
-    "target": "es2018",
+    "target": "es2024",
```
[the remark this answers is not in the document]
khufu-dev: Fixed

.changeset/eleven-months-tickle.md
khufu-dev: Ok bumping major version since next commit will likely have ES2024 code.

packages/core/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
-    "target": "ES2023",
+    "target": "ES2024",
```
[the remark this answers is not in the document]
khufu-dev: This is incorrect.

.changeset/solid-garlics-exist.md
khufu-dev: Based on other discussion and Codex review I'm increasing major version. This will start emitting ES2024 code since
- tsup.config.ts does not set a target, it reads the package tsconfig.json.
- tsconfig.json sets "target": "ES2024"

walton007 (approved):

Issue #1178, opened 2026-04-30
Triage and clear Dependabot vulnerability alerts on default branch


walton007:
## Background

GitHub Dependabot currently reports **70 open vulnerability alerts** on the default branch (`main`):

- 1 critical
- 32 high
- 29 moderate
- 8 low

The repo already uses `pnpm.overrides` as the standard mitigation strategy for transitive dependencies. Most alerts are transitive, two are direct (`esbuild` in `apps/test-server` and `packages/react`).

Full advisory page: <https://github.com/webspatial/webspatial-sdk/security/dependabot>

## Plan

Split the cleanup across multiple PRs to keep risk and review surface small.

### PR 1 (low risk, see #TBD-PR-LINK)

- Bump direct `esbuild` devDependency `^0.23.1 → ^0.25.0` in `apps/test-server/package.json` and `packages/react/package.json`.
- Extend `pnpm.overrides` with patched ranges for transitive packages flagged by Dependabot:
  - `form-data`, `@xmldom/xmldom`, `basic-ftp`, `vite`, `lodash`, `lodash-es`, `path-to-regexp`, `flatted`, `immutable`, `serialize-javascript`, `minimatch`, `rollup`, `@modelcontextprotocol/sdk`, `@remix-run/router`, `react-router`, `validator`, `glob`, `postcss`, `axios`, `follow-redirects`, `brace-expansion`, `picomatch`, `ajv`, `markdown-it`, `mdast-util-to-hast`, `body-parser`, `js-yaml`, `qs`, `diff`, `tmp`, `@eslint/plugin-kit`, `min-document`.
- Constrain the `vite` override to `<7.0.0` so transitive resolution stays on 6.x.

Expected impact: should clear roughly 55 of the 70 alerts (all high / critical that are simple patch / minor bumps).

### PR 2+ (separate evaluations, follow-up)

These require a small compatibility check before bumping:

- `uuid` v3/v5/v6 → v14 (major bump; check direct usages)
- `file-type` → 21.x (major bump in a transitive dev tool)
- `react-router` 7+ / `vite` 7+ (out of scope for the patch sweep above)

## Verification (per PR)

- `pnpm test` (core + react vitest, plus `tsc -p apps/test-server/tsconfig.json`)
- `xcodebuild -project packages/visionOS/web-spatial.xcodeproj -scheme web-spatial -configuration Debug -destination "platform=visionOS Simulator,name=Apple Vision Pro" build`

[...]

PR #1179, opened 2026-04-30
chore(deps): bump esbuild and extend pnpm.overrides to clear Dependabot alerts

walton007:
## Summary

- Bump direct `esbuild` devDependency `^0.23.1 → ^0.25.0` in `apps/test-server/package.json` and `packages/react/package.json` (matches the existing `esbuild` override so the package.json range now reflects what is actually installed).
- Extend root `pnpm.overrides` with patched ranges for the transitive packages currently flagged by Dependabot. See [Tracking issue #1178](https://github.com/webspatial/webspatial-sdk/issues/1178) for the full strategy and the major-version bumps deliberately deferred to follow-up PRs.
- Constrain the `vite` override to `<7.0.0` so transitive resolution stays on the supported 6.x line (otherwise `@vitejs/plugin-react` floats vite to 8.x via peer resolution).

No source code changes; lockfile updated by `pnpm install`.

## Why

Currently 70 open Dependabot alerts on `main` (1 critical, 32 high, 29 moderate, 8 low). Most are transitive and can be cleared without code changes by widening the `pnpm.overrides` map. This PR is the low-risk batch from the plan in #1178.

## Test plan

- [x] `pnpm test` — `@webspatial/core` 110/110, `@webspatial/react-sdk` 80/80, plus `tsc -p apps/test-server/tsconfig.json`
- [x] `xcodebuild -project packages/visionOS/web-spatial.xcodeproj -scheme web-spatial -configuration Debug -destination "platform=visionOS Simulator,name=Apple Vision Pro" build` → `** BUILD SUCCEEDED **`
- [ ] After merge: re-check Dependabot dashboard; remaining alerts will be addressed in follow-up PRs (`uuid` → 14, `file-type` → 21, `react-router` 7+, `vite` 7+).

Refs #1178


JackHFK (approved):

package.json
```diff
@@ -69,8 +69,45 @@
       "prismjs@<1.30.0": ">=1.30.0",
       "@babel/runtime@<7.26.10": ">=7.26.10",
       "@babel/helpers@<7.26.10": ">=7.26.10",
-      "vite@>=6.2.0 <6.2.3": ">=6.2.3",
-      "vite@>=6.2.0 <6.2.4": ">=6.2.4"
+      "vite@>=6.0.0 <6.4.2": ">=6.4.2 <7.0.0",
```
khufu-dev, line 72: Can we just update our dependencies instead of doing overrides. We should be aggressively updating our dependencies since we are at the leading edge of web standards.

khufu-dev (approved):

package.json
walton007: Agree.

PR #1180, opened 2026-04-30
fix: Model ready promise not working

khufu-dev:
When src was not defined on a Model and a property changed the React code would send a blank value for the src parameter to clear out old values on the native side. But if there was no src defined to begin with this would incorrectly push a new source change update and reset the ready promise. When the src or sources property doesn't change the ready promise should not reset.

### DIFF
--- a/packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx
+++ b/packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx
@@ -108,7 +108,7 @@ function SpatializedContent(props: SpatializedStatic3DContentProps) {
     // value needs to be sent to clear the old value
     // TODO: Can native side handle null instead of ''
     spatializedElement.updateProperties({
-      modelURL: modelURL ?? '',
+      modelURL: modelURL ?? (spatializedElement.modelUrl ? '' : modelURL),
       sources,
       autoplay: autoPlay,
       loop,
--- a/.changeset/lucky-drinks-press.md
+++ b/.changeset/lucky-drinks-press.md
@@ -1,0 +1,7 @@
+---
+'web-content': patch
+'@webspatial/react-sdk': patch
+'@webspatial/core-sdk': patch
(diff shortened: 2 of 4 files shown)


packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx
```diff
@@ -108,7 +108,7 @@ function SpatializedContent(props: SpatializedStatic3DContentProps) {
     // value needs to be sent to clear the old value
     // TODO: Can native side handle null instead of ''
     spatializedElement.updateProperties({
-      modelURL: modelURL ?? '',
+      modelURL: modelURL ?? (spatializedElement.modelUrl ? '' : modelURL),
       sources,
```
parthpatelsj, line 112: Is it possible for sources to be []? Do we need similar safeguard for sources?

parthpatelsj (approved):

packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx
khufu-dev: Great question. Fundamentally the problem with partial updates is that the native side cannot distinguish between `undefined`, `null` and value of type `T`. On the native side `undefined` and `null` both get converted to `null`. 
When doing a partial update you might say `updateProps({name})` and then do `updateProps({age})`. Now if you want to do both name and age update you may do `updateProps({name: null, age})`. The native side cannot distinguish between `updateProps({age})` and `updateProps({name: null, age})` both receive name as null. But really in the first case you want `name = undefined` and in the second case it should be `name=null` and should be cleared.
With arrays we don't have this problem because `undefined ≈ undefined`, `[] ≈ null` and `[4, 32] ≈ 4`. The three states can be defined communicated clearly to the native side.

This is why I added a TODO to replace modelUrl with sources while maintaining compatibility for PICO OS.

PR #1183, opened 2026-05-05
feat: Model lazy loading design

khufu-dev:
Updated Model lazy loading technical design. This change moves the lazy loading logic to the web code instead of doing it in native code.

https://claude.ai/code/session_016g7DouZ9bTx5gYUJYA9XnC

### DIFF
--- a/docs/Model.md
+++ b/docs/Model.md
@@ -291,32 +291,31 @@ This feature provides a built-in, intuitive way for users to inspect a 3D model 
 
 - **Gesture Conflict Resolution**: `onSpatialDragStart`, `onSpatialDrag`, and `onSpatialDragEnd` will be disabled when stagemode is set to orbit.
 
-### 8. Lazy Loading (`loading="lazy"`) 🚧
+### 8. Lazy Loading (`loading="lazy"`)
 
-Lazy loading will be managed natively to more accurately determine viewport intersection, removing the need for a web-based `IntersectionObserver` and simplifying the React implementation.
+Match `<img loading="lazy" | "eager">`: when `"lazy"`, the asset isn't fetched until the element is near the viewport; once started, it stays loaded.
 
 #### 8.1. React SDK (`@webspatial/react-sdk`)
 
-- **Prop Forwarding**: The `<Model>` component will accept the `loading?: 'eager' | 'lazy'` prop. This prop will be passed down through `SpatializedStatic3DElementContainer`.
-
-- **No&nbsp;IntersectionObserver**: The previously proposed `useIntersectionLazyLoad` hook and any `IntersectionObserver` logic will be removed entirely from the React SDK for this feature. The component's responsibility is simply to inform the native layer of the desired loading behavior.
+In `SpatializedStatic3DElementContainer.tsx`, derive `effectiveLoading` from an `IntersectionObserver` on the portal element and forward it via the existing `updateProperties({ … })` call. Initial value is `'lazy'` only when `props.loading === 'lazy'`; once it flips to `'eager'`, it sticks (the observer disconnects). But if the sources change and the loading property is `lazy` then the `IntersectionObserver` is re-attached and `updateProperties({ … })` is called with `loading: lazy`. Lazy loading a Model within a spatialized element is intrinsically handled by this implementation because the root web page contains hidden DOM elements for all spatialized elements.
 
 #### 8.2. Core SDK (`@webspatial/core-sdk`)
 
-- **JSB Command Update**: The `UpdateSpatializedStatic3DElementProperties` command in `JSBCommand.ts` will be extended to include `loading?: 'eager' | 'lazy'`. This value is forwarded directly to the native layer.
+In `SpatialSession.ts` update `createSpatializedStatic3DElement` to accept the `loading` property in addition to `src` and `sources`, which defaults to `'eager'`.
+Extend `CreateSpatializedStatic3DElementCommand` and `UpdateSpatializedStatic3DElementProperties` in `JSBCommand.ts` with `loading?: 'eager' | 'lazy'`, forwarded directly to native.
 
 #### 8.3. Native visionOS Layer (`packages/visionOS`)
 
-- **Native Intersection Detection**: The native layer will now be responsible for calculating if the element is visible within the webview's viewport.
-  1.  **State Management**: `SpatializedStatic3DElement.swift` will store the `loading` mode and a new computed property `var isIntersecting: Bool`.
-  2.  **Viewport & Scroll Data**: To calculate intersection, we need context from the webview.
-      - `SpatialScene.spatialWebViewModel` already exposes `scrollOffset` via its `addScrollUpdateListener`.
-      - We will extend `SpatialWebViewModel.swift` to also expose the webview's viewport size. A new method or computed property will be added: `var viewportSize: CGSize { return webview?.scrollView.frame.size ?? .zero }`.
-  3.  **Intersection Calculation**: With the element's 2D frame (`clientX`, `clientY`, `width`, `height`), the `scrollOffset`, and the `viewportSize`, the native `SpatializedStatic3DElement` can accurately compute if its frame overlaps with the visible area of the webview at any time. This calculation will be triggered whenever a scroll update occurs.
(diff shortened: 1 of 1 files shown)


khufu-dev:
@codex Review again

docs/Model.md
```diff
@@ -291,32 +291,31 @@ This feature provides a built-in, intuitive way for users to inspect a 3D model
...
 
-Lazy loading will be managed natively to more accurately determine viewport intersection, removing the need for a web-based `IntersectionObserver` and simplifying the React implementation.
+Match `<img loading="lazy" | "eager">`: when `"lazy"`, the asset isn't fetched until the element is near the viewport; once started, it stays loaded.
 
 #### 8.1. React SDK (`@webspatial/react-sdk`)
 
-- **Prop Forwarding**: The `<Model>` component will accept the `loading?: 'eager' | 'lazy'` prop. This prop will be passed down through `SpatializedStatic3DElementContainer`.
-
-- **No&nbsp;IntersectionObserver**: The previously proposed `useIntersectionLazyLoad` hook and any `IntersectionObserver` logic will be removed entirely from the React SDK for this feature. The component's responsibility is simply to inform the native layer of the desired loading behavior.
+In `SpatializedStatic3DElementContainer.tsx`, derive `effectiveLoading` from an `IntersectionObserver` on the portal element and forward it via the existing `updateProperties({ … })` call. Initial value is `'lazy'` only when `props.loading === 'lazy'`; once it flips to `'eager'`, it sticks (the observer disconnects).
```
parthpatelsj: Imagine two <Model>s:
A: `<Model loading="eager" ...>`
B: `<Model loading="lazy" ...>` –

Once B has flipped and started loading, does it behave exactly like A from that point onward?

i.e. autoPlay starts at the same moment relative to onLoad as in eager

parthpatelsj: How does this behave with nested scroll containers?
i.e. something like....
```
<!-- outer page scroll -->
  <div style="height: 2000px; overflow-y: auto">
    Some content...
    <div style="height: 500px; overflow-y: auto">
      <!-- inner scrollable panel -->
      <Model loading="lazy" src="/modelasset/cone.glb" enable-xr />
    </div>
  </div>
```

For nested scroll containers, is lazy loading checked against the main window viewport or the inner scroll container?

khufu-dev: Yes once it has flipped to "eager" it behaves exactly like A or as if `loading="lazy"` was not set.

docs/Model.md
```diff
@@ -291,32 +291,31 @@ This feature provides a built-in, intuitive way for users to inspect a 3D model
 
 - **Gesture Conflict Resolution**: `onSpatialDragStart`, `onSpatialDrag`, and `onSpatialDragEnd` will be disabled when stagemode is set to orbit.
 
-### 8. Lazy Loading (`loading="lazy"`) 🚧
+### 8. Lazy Loading (`loading="lazy"`)
 
-Lazy loading will be managed natively to more accurately determine viewport intersection, removing the need for a web-based `IntersectionObserver` and simplifying the React implementation.
+Match `<img loading="lazy" | "eager">`: when `"lazy"`, the asset isn't fetched until the element is near the viewport; once started, it stays loaded.
```
parthpatelsj, line 296: What happens if the app flips loading from 'lazy' to 'eager' (or vice versa) after mount?

docs/Model.md
khufu-dev: It will be checked against the inner scroll container. This is one of the benefits of doing lazy loading on the web side using IntersectionObserver instead of native.

parthpatelsj: In this case, will lazy loading be supported in spatial div's that are not in view?

parthpatelsj (approved):

docs/Model.md
khufu-dev: This proposal is only for static 3D model and will not affect spatial divs and their lazy loading behaviour.

docs/Model.md
khufu-dev: This question gets into the nuances of the HTML specs.
1. When loading switches from `lazy` to `eager` the model will start downloading immediately.
2. When loading switches from `eager` to `lazy` the loading will not be interrupted.
3. When loading is `lazy`, the model comes into view, then goes out of view and then the sources change. In this case new sources are loaded lazily when the model comes into view again.

https://html.spec.whatwg.org/multipage/images.html#updating-the-image-data
https://html.spec.whatwg.org/multipage/urls-and-fetching.html#lazy-loading-attributes

parthpatelsj: Makes sense, so it will be in parity with html spec

docs/Model.md
parthpatelsj: Sorry, I should have been more clear. But, it was more so if you have a model in a spatial div. Then, because spatial div doesn't know whether it's in view or not, it will always load the model.

docs/Model.md
```diff
@@ -291,32 +291,31 @@ This feature provides a built-in, intuitive way for users to inspect a 3D model
...
-      - If `shouldLoad` is `false`, the view will render the poster/back-plane or a lightweight placeholder.
-      - The model loading process (`Model3D(url:)`) will only be initiated when `shouldLoad` first becomes `true`.
-  5.  **Scroll Updates**: On each scroll event received from `addScrollUpdateListener`, the `isIntersecting` state for all lazy elements will be re-evaluated. If an element's `isIntersecting` status changes from `false` to `true`, and it hasn't loaded yet, the loading process is triggered.
+Add `var loading: String = "eager"` to `SpatializedStatic3DElement.swift` and gate `allSources` on it:
+
+```swift
+var allSources: [ModelSource] {
+    guard loading == "eager" else { return [] }
+    return if let modelURL { [ModelSource(src: modelURL, type: nil)] + sources } else { sources }
+}
```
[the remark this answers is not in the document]
khufu-dev: Yes this obviously has to be added.

docs/Model.md
```diff
@@ -291,32 +291,31 @@ This feature provides a built-in, intuitive way for users to inspect a 3D model
...
+```
+
+`SpatializedStatic3DView.swift` is untouched. While lazy, `allSources` is `[]`, the existing `if !allSources.isEmpty` gate falls through to `EmptyView()`, and nothing renders, fetches, or autoplays. Poster and autoplay defer for free — both are only reached via `.task(id: allSources)`, which doesn't fire until the eager transition.
```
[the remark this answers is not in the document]
khufu-dev: Fixed, updated proposal to mention changes need to be done to SpatializedStatic3DView.swift

docs/Model.md
```diff
@@ -291,32 +291,31 @@ This feature provides a built-in, intuitive way for users to inspect a 3D model
...
```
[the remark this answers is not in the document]
khufu-dev: Fixed, updated proposal to mention changes need to be done to SpatializedStatic3DView.swift

docs/Model.md
khufu-dev: Great question and I had to verify this. WebSpatial keeps a hidden DOM element of all spatial elements and models in the root WebPage with their CSS position. As long as I use IntersectionObserver on this hidden DOM element it should detect when the Model comes into view even it's inside a spatialized container.

https://github.com/user-attachments/assets/cfb36a7b-55dd-4988-9c0c-cd9735b5cd48

docs/Model.md
```diff
@@ -291,32 +291,31 @@ This feature provides a built-in, intuitive way for users to inspect a 3D model
...
 
 #### 8.1. React SDK (`@webspatial/react-sdk`)
 
-- **Prop Forwarding**: The `<Model>` component will accept the `loading?: 'eager' | 'lazy'` prop. This prop will be passed down through `SpatializedStatic3DElementContainer`.
-
-- **No&nbsp;IntersectionObserver**: The previously proposed `useIntersectionLazyLoad` hook and any `IntersectionObserver` logic will be removed entirely from the React SDK for this feature. The component's responsibility is simply to inform the native layer of the desired loading behavior.
+In `SpatializedStatic3DElementContainer.tsx`, derive `effectiveLoading` from an `IntersectionObserver` on the portal element and forward it via the existing `updateProperties({ … })` call. Initial value is `'lazy'` only when `props.loading === 'lazy'`; once it flips to `'eager'`, it sticks (the observer disconnects). But if the sources change and the loading property is `lazy` then the `IntersectionObserver` is re-attached and `updateProperties({ … })` is called with `loading: lazy`. Lazy loading a Model within a spatialized element is intrinsically handled by this implementation because the root web page contains hidden DOM elements for all spatialized elements.
```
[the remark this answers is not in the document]
khufu-dev: I think this is an acceptable compromise for now

PR #1191, opened 2026-05-08
fix: incorrect load/error callbacks on Model

khufu-dev:
The previous wiring routed load callbacks through .onAppear modifiers inside the view body, which caused three problems:

- onError fired on the first non-empty render before the .task had a chance to flip isLoading, because (asset == nil && !isLoading) was indistinguishable from "load failed".
- onLoad was attached to Model3D's content closure, so it could fire on re-renders or be skipped when SwiftUI reused the view.
- The (idle vs failed) ambiguity blocked any future lazy-loading gate.

Replace the three @State vars with a LoadState enum (idle / loading / loaded / failed) and fire onLoad/onError exactly once from inside loadSources(). Hoist the .task(id:) to the outer Group so empty <-> non-empty transitions reset cleanly, and guard on Task.isCancelled so an in-flight source swap doesn't emit a spurious failure.

https://claude.ai/code/session_01VjxxinriXVX19ULdjhzRt1

### DIFF
--- a/.changeset/purple-phones-cry.md
+++ b/.changeset/purple-phones-cry.md
@@ -1,0 +1,7 @@
+---
+'@webspatial/platform-visionos': patch
+'web-content': patch
+'ci-test': patch
+---
+
+fix: incorrect load/error callbacks on Model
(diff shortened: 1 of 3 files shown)


khufu-dev:
@codex does this resolve https://github.com/webspatial/webspatial-sdk/issues/1098?

parthpatelsj (approved):

Issue #1220, opened 2026-05-14
Doing `enable-xr` on <Reality> crashes the page


parthpatelsj:

parthpatelsj:
@codex Can you fix this bug? Task: Investigate and fix WebSpatial issue #1220 — calling `enable-xr` on a `<Reality>` element crashes the page.

parthpatelsj:
Issue is fixed with PR #1222

PR #1222, opened 2026-05-15
Fix Reality enable-xr crash

parthpatelsj:
### Motivation
- Prevent a runtime crash that happened when JSX transformed `<Reality enable-xr>` into a 2D spatialized container and forwarded the raw `enable-xr` attr to the host DOM node.
- Ensure `Reality` accepts and strips `enable-xr` without being rewritten by the JSX helper so Reality remains a proper 3D host.

### Description
- Treat `Reality` like `Model` in the JSX runtime by updating `packages/react/src/jsx/jsx-shared.ts` to avoid rewriting `Reality` into a 2D spatialized container.
- Add explicit `enable-xr` support to `RealityProps` and destructure/strip the prop in `packages/react/src/reality/components/Reality.tsx` so it is not forwarded to the host DOM element, and add `Reality.displayName` for clarity.
- Add a regression test `packages/react/src/jsx/jsx-shared.test.tsx` that verifies the JSX transform does not rewrite `Reality` and that `enable-xr` is stripped before forwarding.
- Update `packages/react/vitest.config.ts` to alias workspace source paths for `@webspatial/core-sdk` and `@webspatial/react-sdk` so tests can import runtime internals from source during the test run.

### Testing
- Ran `pnpm --filter @webspatial/react-sdk test`; all package tests passed (react package: 96 tests passed).
- Ran full workspace suite via `pnpm test`; all workspace tests completed successfully (core + react + other workspace tests passed).

Related to #1220 
------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a0663075470832397af4d948a0132a3)

### DIFF
--- a/packages/react/src/reality/components/Reality.tsx
+++ b/packages/react/src/reality/components/Reality.tsx
@@ -147,4 +147,3 @@ export const Reality = forwardRef<SpatializedElementRef, RealityProps>(
     )
   },
 )
-Reality.displayName = 'Reality'
--- a/packages/react/src/reality/components/Reality.tsx
+++ b/packages/react/src/reality/components/Reality.tsx
@@ -18,13 +18,12 @@ import { useRealityEvents } from '../hooks'
 
 export type RealityProps = Omit<
   React.ComponentPropsWithRef<'div'>,
-  'onSpatialContentReady' | 'enable-xr'
-> & {
-  'enable-xr'?: boolean
-} & EntityEventHandler
+  'onSpatialContentReady'
+> &
+  EntityEventHandler
 
 export const Reality = forwardRef<SpatializedElementRef, RealityProps>(
-  function RealityBase({ children, 'enable-xr': _enableXR, ...inProps }, ref) {
+  function RealityBase(props, ref) {
     const insideAttachment = useInsideAttachment()
     if (insideAttachment) {
       console.warn(
@@ -33,6 +32,8 @@ export const Reality = forwardRef<SpatializedElementRef, RealityProps>(
       return null
     }
     const {
+      children,
+      'enable-xr': _enableXR,
       onSpatialTap,
       onSpatialDragStart,
       onSpatialDrag,
@@ -41,8 +42,8 @@ export const Reality = forwardRef<SpatializedElementRef, RealityProps>(
       onSpatialRotateEnd,
       onSpatialMagnify,
       onSpatialMagnifyEnd,
-      ...props
-    } = inProps
+      ...divProps
+    } = props as Omit<RealityProps, 'ref'> & { 'enable-xr'?: boolean }
     const ctxRef = useRef<RealityContextValue | null>(null)
 
     const creationId = useRef(0)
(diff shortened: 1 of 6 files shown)


walton007:
@codex can you help CR?

packages/react/src/reality/components/Reality.tsx
```diff
@@ -18,12 +18,13 @@ import { useRealityEvents } from '../hooks'
 
 export type RealityProps = Omit<
   React.ComponentPropsWithRef<'div'>,
-  'onSpatialContentReady'
-> &
-  EntityEventHandler
+  'onSpatialContentReady' | 'enable-xr'
+> & {
+  'enable-xr'?: boolean
```
walton007: I think the jsx-shared.ts part of this fix is correct: Reality should not be rewritten into a 2D spatialized container just because enable-xr is present.

However, the Reality.tsx type change does not look right. We do not intend Reality to publicly support the enable-xr prop, so RealityProps should not include 'enable-xr'?: boolean. The fix should keep the JSX runtime special case while leaving RealityProps without enable-xr.

parthpatelsj: @codex Can you leave Reality.tsx alone (we don't want Reality to publicly support enable-xr), and keep the fix jsx runtime special case.

packages/react/src/reality/components/Reality.tsx
```diff
@@ -138,10 +140,11 @@ export const Reality = forwardRef<SpatializedElementRef, RealityProps>(
...
           spatializedContent={content}
-          {...props}
+          {...divProps}
         />
         {isReady && children}
       </RealityContext.Provider>
     )
   },
 )
+Reality.displayName = 'Reality'
```
khufu-dev: Unrelated to bug fix.
```suggestion
```

packages/react/vitest.config.ts
```diff
@@ -1,8 +1,17 @@
...
 export default defineConfig({
   plugins: [react()],
+  resolve: {
+    alias: {
+      '@webspatial/core-sdk': fileURLToPath(
+        new URL('../core/src', import.meta.url),
+      ),
+      '@webspatial/react-sdk': fileURLToPath(new URL('./src', import.meta.url)),
+    },
+  },
```
khufu-dev: Avoid this change, instead fix the tests to work with the existing setup.
```suggestion
```

packages/react/src/reality/components/Reality.tsx
```diff
@@ -40,8 +42,8 @@ export const Reality = forwardRef<SpatializedElementRef, RealityProps>(
       onSpatialRotateEnd,
       onSpatialMagnify,
       onSpatialMagnifyEnd,
-      ...props
-    } = inProps
+      ...divProps
+    } = props as Omit<RealityProps, 'ref'> & { 'enable-xr'?: boolean }
```
khufu-dev: Avoid the cast. Can you just turn off the enable-xr in the SpatializedContainer instead of reading it from the props? It should be ignored.

```suggestion
    } = props
```

khufu-dev (changes requested):

packages/react/vitest.config.ts
```diff
@@ -1,3 +1,4 @@
+import { fileURLToPath } from 'node:url'
```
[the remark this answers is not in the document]
parthpatelsj: Unrelated to this PR

khufu-dev (approved):

PR #1233, opened 2026-05-25
feat(react-sdk)!: lazy-load default entry + eager distribution

walton007:
## Summary

Implements OpenSpec change **`lazy-load-spatial-runtime`**: replaces the old dual-build (`dist/web` vs `dist/default`) + `@webspatial/vite-plugin` alias model with a **runtime lazy-load default entry** and an optional **`@webspatial/react-sdk/eager`** entry for spatial-only apps.

At a high level:

- **Default entry** (`@webspatial/react-sdk`): ships facades, hook placeholders, `bootSpatial()`, and the runtime bridge. The real spatial implementation lives in a separate `./spatial` chunk and is loaded only after `bootSpatial()` in a WebSpatial runtime. Plain web / SSR use documented per-component fallbacks and do not fetch the spatial chunk.
- **Eager entry** (`@webspatial/react-sdk/eager`): statically links spatial implementations for CSR-only spatial surfaces. `bootSpatial()` is an immediate no-op stub. There is no lazy size cap for eager.
- **Breaking surface**: removes `/web` and `/default` subpaths, removes internal container / monitor exports from the public surface, deprecates `createElement`, and makes `@webspatial/vite-plugin` unnecessary for ordinary consumers.
- **Size contract**: the lazy default entry keeps the typical `import { Model, bootSpatial }` marginal delta at **≤ 5 KB gzipped** via `tests/marginal-delta-vite`.

## Reviewer Start Here

If you only have limited review time, read in this order:

1. **Code change map** below to understand the PR layout.
2. Runtime activation path: `packages/react/src/runtime/`.
3. Default-entry facades and fallbacks: `packages/react/src/facades/`, `packages/react/src/hooks/`, `packages/react/src/index.ts`.
4. JSX marker handling: `packages/react/src/jsx/` and `packages/react/src/internal/facades-client.ts`.
5. Consumer-shaped fixtures: `apps/spatial-vite-min`, `apps/spatial-next-min`, `apps/spatial-next-eager-min`, `apps/spatial-remix-min`, `apps/spatial-rspack-min`.
6. Spec disputes / exact contracts: `openspec/changes/lazy-load-spatial-runtime/specs/spatial-lazy-load/spec.md`.

## Code Change Map

| Area | Main paths | What changed / what to review |
| --- | --- | --- |
| Package exports + build shape | `packages/react/package.json`, `packages/react/tsup.config.ts`, `packages/react/src/index.ts`, `packages/react/src/eager.ts`, `packages/react/src/spatial/index.ts` | Defines the lazy default entry, eager entry, spatial subpath, JSX runtime subpaths, and chunking rules. Review this for public API shape and bundler compatibility. |
| Runtime bridge + boot | `packages/react/src/runtime/` | Adds the bridge state machine, `bootSpatial()`, `isSpatialReady()`, load-error listeners, `SpatialBoot`, eager stubs, and runtime errors. Review retry/error/SSR/no-runtime behavior here. |
| Facades + hooks | `packages/react/src/facades/`, `packages/react/src/hooks/`, `packages/react/src/runtime/useSpatialReady.ts` | Default entry renders fixed per-component fallbacks before boot and delegates to real spatial implementations after readiness. `useSpatialReady()` is the app wrapper escape hatch; no generic per-facade `fallback` prop in v1. |
| JSX runtime | `packages/react/src/jsx/`, `packages/react/src/internal/facades-client.ts` | Unified JSX runtime strips spatial marker props and wraps marker elements through facade HOCs. This is where `<div enable-xr>` and `<model>` marker behavior is preserved across lazy/eager/default paths. |
| Core/runtime fail-fast cleanup | `packages/core/src/`, `packages/react/src/runtime/convertCoordinate.ts` | Tightens unsupported-runtime behavior: SSR platform misuse and unsupported `convertCoordinate()` now fail fast instead of silently degrading. |
| Consumer fixtures | `apps/spatial-vite-min`, `apps/spatial-next-min`, `apps/spatial-next-eager-min`, `apps/spatial-remix-min`, `apps/spatial-rspack-min`, `apps/_shared/webspatial-request.ts` | Adds consumer-shaped Vite / Next / eager Next / React Router / Rspack fixtures and shared request classification. These validate package exports, SSR boundaries, lazy boot, eager CSR, and bundler compatibility. |
| Test server + demo cleanup | `apps/test-server/`, `tests/autoTest/` | Moves the internal dev site toward the lazy default path, removes legacy `XR_ENV` defines, and keeps WebSpatial-heavy regression demos working. |
| CI + validation | `.github/workflows/ci.yml`, `tests/marginal-delta-vite`, package tests | Adds fixture build coverage and size-budget validation so lazy/eager fixture regressions are visible in CI. |
| Specs / docs / release notes | `openspec/changes/lazy-load-spatial-runtime/`, `docs/migration/lazy-load-spatial-runtime.md`, `docs/react-sdk-product-alignment.md`, `.changeset/` | Normative contracts, reviewer guide, migration story, and release notes for the breaking API changes. |

## Key Contracts To Check

- **Default entry is web-first**: plain web and SSR render facade fallbacks without loading the spatial chunk.
- **`bootSpatial()` is the single activation path** for loading real spatial implementations in WebSpatial runtimes.
- **`SpatialBoot` v1 surface is small**: `children`, `onReady`, `onError`; no public `gate` / `fallback` props. It renders `null` until boot succeeds.
- **Eager is CSR-only for spatial primitives**: use the default entry for SSR HTML that needs facade fallbacks, or CSR-gate the eager subtree.
- **No default/eager mixing in one bundle**: consumers should pick one React SDK import root per bundle.
- **JSX runtime stays entry-agnostic**: marker stripping / wrapping must not depend on whether the consumer imports default or eager components.
- **Size budget applies only to lazy default**: eager intentionally inlines spatial implementations and has no separate 5 KB cap.

## Review Entry Points

| Doc | Purpose |
| --- | --- |
| [`REVIEW.md`](https://github.com/webspatial/webspatial-sdk/blob/feat/eager-mode-entry/openspec/changes/lazy-load-spatial-runtime/REVIEW.md) | Deep reviewer guide: requirements index, parity audit, roadmap. |
| [`tasks.md`](https://github.com/webspatial/webspatial-sdk/blob/feat/eager-mode-entry/openspec/changes/lazy-load-spatial-runtime/tasks.md) | Implementation checklist and known follow-ups. |
| [`spec.md`](https://github.com/webspatial/webspatial-sdk/blob/feat/eager-mode-entry/openspec/changes/lazy-load-spatial-runtime/specs/spatial-lazy-load/spec.md) | Normative source of truth when code/spec questions come up. |
| [`Migration guide`](https://github.com/webspatial/webspatial-sdk/blob/feat/eager-mode-entry/docs/migration/lazy-load-spatial-runtime.md) | Consumer upgrade recipes and breaking-change guidance. |

## Out Of Scope / Non-Blocking Follow-Ups

- [ ] §11 — cross-repo `@webspatial/vite-plugin` deprecation tracking.
- [ ] §12.1–12.3 — test-server / autoTest / ci-test full **dist** + chunk-fetch assertions.
- [ ] §12.7–12.8 — Turbopack / Module Federation validation.
- [ ] §15.8 — promote remaining parity `it.todo` cases into asserted spec coverage.

## Test Plan

- [x] `openspec validate lazy-load-spatial-runtime --strict`
- [x] `pnpm --filter @webspatial/react-sdk test` (351 tests; 4 tracked `it.todo`)
- [x] `pnpm run test:size-budget`
- [x] `pnpm run test:rspack-compat`
- [x] Fixture builds covered by `.github/workflows/ci.yml`
- [ ] `npm run test:auto` / `npm run ciTest` (environment-dependent)

## Changesets

Includes `.changeset/` entries for the lazy-load slices, eager entry, breaking API surface changes, SSR/runtime fail-fast updates, fixture/docs changes, and migration notes.

Made with [Cursor](https://cursor.com)


packages/react/src/facades/Model.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * structurally identical to the real-impl Path 2 unsupported branch in
+ * `src/Model.tsx` (the `enable-xr={false}` / `!Spatial.runInSpatialWeb()`
+ * gate), per `runtime-capabilities` "`Model` exception fallback"
+ * Scenario. Verified by `src/__tests__/parity.test.tsx` ("Model parity"
+ * suite); changes to either path MUST update both paths or the parity
+ * test will fail.
+ */
+function ModelFacadeImpl(props: ModelProps, ref: ForwardedRef<ModelRef>) {
+  const ready = useSpatialReady()
+  if (!ready) {
```
parthpatelsj, line 24: If developer has code like this: 

```
function MyDapp() {
  const modelRef = useRef(null);

  useEffect(() => {
    // 1. App mounts. "ready" is false.
    // modelRef.current points to the fallback <model> element.
    const el = modelRef.current;

    const testClick = () => console.log("Model clicked!");
    el.addEventListener('click', testClick);

    return () => el.removeEventListener('click', testClick);
  }, []); // Only runs once on mount

  return <Model ref={modelRef} src="drone.glb" />;
}
```

This would fail right? In beginning, spatial engine isn't ready so it would render fallback `renderModelFallback`, 

When `useEffect` runs, it finds <model> and attach testClick event 

Then, lets say `bootSpatial()` ready is true....then when swap happens <RealModel> 

React will delete old fallback, and mount new element. 

But, the click listener is still attached to the old, deleted <model> element sitting dead in the browser's memory.

parthpatelsj:
@codex Do full code review for this feature.

packages/react/src/facades/Model.tsx
walton007: Agreed. With lazy-load facades, a one-shot useEffect([]) that binds to modelRef.current on first mount attaches to the fallback <model>. After bootSpatial() resolves, the facade swaps to the real implementation; the parent effect doesn't re-run, so listeners stay on detached DOM. Fix: use <SpatialBoot> / await bootSpatial() before mounting spatial UI, or depend on useSpatialReady() (not []) when boot is late. We'll document this in the migration guide.

walton007: This is a real caveat for late-boot facades. If `Model` mounts before `bootSpatial()` resolves, the forwarded ref can point at the fallback `<model>` for the first commit; when the facade swaps to the real implementation, a parent `useEffect([])` does not re-run, so imperative listeners captured from the fallback node will not move to the real implementation.

I’m treating this as an integration contract rather than changing facade swap semantics: `<SpatialBoot>` is the recommended path when spatial UI needs imperative refs/listeners, because children mount only after the real implementation is ready. For custom late-boot wiring, the effect should depend on `useSpatialReady()` and treat fallback vs real nodes as separate lifetimes.

Documented in `docs/migration/lazy-load-spatial-runtime.md` under “Imperative refs during a late boot” in commit `b84c5e5a`.

.changeset/react-sdk-lazy-load-major.md
```diff
@@ -0,0 +1,16 @@
+---
+'@webspatial/react-sdk': major
+---
+
+BREAKING: replace the React SDK dual-build distribution with a web-first lazy-load default entry and a CSR-only eager entry.
+
+- The default `@webspatial/react-sdk` entry now ships facades, hook placeholders, `bootSpatial()`, `isSpatialReady()`, `useSpatialReady()`, `onSpatialLoadError()`, `WebSpatialBootError`, and stateless utility exports. Real spatial implementations live behind the dynamically loaded `@webspatial/react-sdk/spatial` chunk and are activated through `bootSpatial()`.
```
khufu-dev: This is a major breaking change. Keep the summary short and explain to the user what changes need to be made so they can upgrade from v1 to v2. Internal details can be exposed in other documentation.

apps/test-server/src/pages/reality/empty.tsx
```diff
@@ -64,7 +64,7 @@ function App() {
       <button
         className={btnCls}
         onClick={() => {
-          location.href = './index.html'
+          location.href = '/#/reality'
```
khufu-dev: Is this change required?

apps/test-server/index.tsx
```diff
@@ -275,7 +276,13 @@ const init = () => {
 
   ReactDOM.createRoot(rootElement).render(
     <React.StrictMode>
-      <App />
+      <SpatialBoot
+        onError={(err: WebSpatialBootError) => {
+          console.error('[test-server] bootSpatial failed', err)
```
khufu-dev: Realistically what is the user supposed to do when there's an error? Do we show them an error on the HTML page in DEV mode.

apps/spatial-next-eager-min/package.json
```diff
@@ -0,0 +1,26 @@
+{
```
khufu-dev, line 1: I think it's better to have these spatial-* test app in a separate test repository. This way we can test across versions similar to our end developers. When we bundle these end to end tests in our code repo we cannot detect unexpected breaking changes.

apps/test-server/src/pages/spatialStyleTest/CubeComponent.tsx
```diff
@@ -1,5 +1,4 @@
-import { useState } from 'react'
-import { CSSProperties } from 'styled-components'
+import { useState, type CSSProperties } from 'react'
```
khufu-dev, line 1: This change seems unrelated to this feature.

docs/migration/lazy-load-spatial-runtime.md
```diff
@@ -0,0 +1,379 @@
...
+```ts
+// vite.config.ts (1.5.x)
+import { defineConfig } from 'vite'
+import react from '@vitejs/plugin-react'
+import webSpatial from '@webspatial/vite-plugin'
+
+export default defineConfig({
+  plugins: [react(), webSpatial({ output: 'avp' })],
+})
+```
```
khufu-dev: Do not document this. Our public documentation already removed references to `@webspatial/vite-plugin`. This will confuse developers and AI.

docs/migration/lazy-load-spatial-runtime.md
```diff
@@ -0,0 +1,379 @@
...
+```tsx
+import { SpatialBoot, WebSpatialBootError } from '@webspatial/react-sdk'
+
+export function AppRoot() {
+  return (
+    <SpatialBoot
+      onError={(err: WebSpatialBootError) => {
+        // Show error UI; retry with bootSpatial() if needed.
+        console.error(err)
+      }}
```
khufu-dev: I don't think this is useful for existing developers.

docs/migration/lazy-load-spatial-runtime.md
```diff
@@ -0,0 +1,379 @@
...
+- It is **idempotent**: concurrent / repeated calls share the same promise.
+- It supports **retry on failure**: after a rejection, calling `bootSpatial()` again starts a fresh `import()`.
+- It **never throws raw `import()` errors**; rejection is always a `WebSpatialBootError` whose `cause` is the underlying error.
+
+If you call `bootSpatial()` without `<SpatialBoot>`, facades and hook placeholders keep rendering their fallback until the promise resolves; on the next React commit after the bridge becomes ready, **mounted facades automatically swap to the real implementation** (no `key` change required).
+
+### Imperative refs during a late boot
+
+If a facade mounts before `bootSpatial()` resolves, refs point at the fallback output for that first commit. For `Model`, that means a forwarded ref can initially point at the degraded native `<model>` fallback. When the spatial chunk becomes ready, the facade replaces that fallback with the real `Model` implementation; React does not re-run a parent `useEffect([])` just because `ref.current` changed.
```
khufu-dev: This is problematic because `modelRef.current.ready` will now be broken and developers will not know when they can execute other APIs such as `entityTransform`

khufu-dev (changes requested):
We already removed using @webspatial/vite-plugin for the end user but your overview has multiple references to this. Make sure not to reference @webspatial/vite-plugin in your changeset or documentation otherwise it confuses the AI.

packages/visionOS/web-spatial/manifest.swift
```diff
@@ -6,8 +6,8 @@ var pwaManager = PWAManager()
 struct PWAManager: Codable {
     var isLocal: Bool = false
 
-//    var start_url: String = "http://localhost:5173/#/"
-    var start_url: String = "http://localhost:5173/#/runtime-capabilities"
+    var start_url: String = "http://localhost:3050"
+//    var start_url: String = "http://localhost:3030/lazy"
```
khufu-dev: Don't commit this change.

packages/react/src/facades/Model.tsx
```diff
@@ -0,0 +1,69 @@
...
+    onSpatialRotateEnd: _onSpatialRotateEnd,
+    onSpatialMagnify: _onSpatialMagnify,
+    onSpatialMagnifyEnd: _onSpatialMagnifyEnd,
+    spatialEventOptions: _spatialEventOptions,
+    'enable-xr': _enableXR,
+    ...modelProps
+  } = props
+  // Native <model> is a non-React intrinsic; createElement bypasses the
+  // JSX intrinsic-type check while still preserving ref forwarding.
+  return createElement('model', { ...modelProps, ref })
```
khufu-dev: Use normal JSX 
```suggestion
  return <model ref={ref} {...modelProps} />
```

packages/react/src/facades/Model.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return <RealModel {...props} ref={ref} />
+}
+
+export const Model = forwardRef<ModelRef, ModelProps>(ModelFacadeImpl)
+Model.displayName = 'Model'
+// Brand so the JSX runtime short-circuits `Model` regardless of whether the
+// default-entry facade or the eager-entry real implementation reaches it.
+markWebSpatialPrimitive(Model, 'Model')
+
+function renderModelFallback(
```
khufu-dev, line 38: Make this a normal React component. AI is struggling with the non standard `model` element.

packages/react/src/facades/Model.tsx
```diff
@@ -0,0 +1,69 @@
...
+ * renders degraded `<model>` tag" Scenarios. Not wrapped in
+ * `React.memo` (per facade conventions).
+ *
+ * **PARITY (spec tasks.md §15.6)**: this Path 1 fallback MUST stay
+ * structurally identical to the real-impl Path 2 unsupported branch in
+ * `src/Model.tsx` (the `enable-xr={false}` / `!Spatial.runInSpatialWeb()`
+ * gate), per `runtime-capabilities` "`Model` exception fallback"
+ * Scenario. Verified by `src/__tests__/parity.test.tsx` ("Model parity"
+ * suite); changes to either path MUST update both paths or the parity
+ * test will fail.
```
khufu-dev: This comment is not helpful.
```suggestion
```

packages/react/src/runtime/bridge.ts
```diff
@@ -0,0 +1,171 @@
...
+}
+
+function toBootError(error: unknown, attempt: number): WebSpatialBootError {
+  if (error instanceof WebSpatialBootError) {
+    return error
+  }
+  return new WebSpatialBootError({ cause: error, attempt })
+}
+
+export function getSpatialImpl(): SpatialImplementation | null {
```
khufu-dev, line 57: I think there should be a version of this function that returns a not null value. So that children of SpatialBoot don't have to force unwrap.

packages/react/src/facades/Model.tsx
```diff
@@ -0,0 +1,69 @@
...
+ */
+function ModelFacadeImpl(props: ModelProps, ref: ForwardedRef<ModelRef>) {
+  const ready = useSpatialReady()
```
khufu-dev, line 23: I believe when this changes from `false` to `true` the `ref.current.ready` will reference the old `<model>` component and trigger an incorrect success or failure.

JackHFK (approved):

.changeset/react-sdk-lazy-load-major.md
walton007: Updated in `3dbd09a6`.

The major changeset now stays focused on the v1 -> v2 upgrade path: choose the default vs eager entry, replace removed `/web` and `/default` imports, gate spatial UI with `<SpatialBoot>` / `bootSpatial()` / `useSpatialReady()` before runtime-only APIs or imperative refs, and update code that used removed implementation-detail exports. Internal implementation details were removed from the release-note summary and left to the migration guide.

apps/test-server/src/pages/reality/empty.tsx
walton007: Reverted this navigation cleanup because it is not required for the lazy-load entry change. We can handle the test-server back-button route separately if needed.

apps/spatial-next-eager-min/package.json
walton007: I agree that a separate consumer test repository is the better long-term place for true cross-
  version compatibility testing. It would let us install published package versions the same way
  end developers do and catch breakages that a workspace-linked monorepo fixture cannot catch.

  For this PR, I’d like to keep these `spatial-*` apps as in-repo smoke fixtures because they
  validate the new export shape, SSR/CSR boundaries, lazy vs eager entry behavior, and bundler
  compatibility against the current workspace changes before publish. They are not intended to
  replace external cross-version tests.

  I’ll track the separate consumer test repo as a follow-up so we can cover published-version
  upgrade scenarios independently.

apps/test-server/src/pages/spatialStyleTest/CubeComponent.tsx
walton007: This is a small type-only fix needed by the test-server build after the lazy-load test-server
  updates.

  The previous import used `CSSProperties` from `styled-components`, which is not the same type as
  React’s `style` prop type here and causes a TypeScript mismatch for `<div style={cubeStyle as
  CSSProperties}>`. Switching to `React.CSSProperties` keeps the runtime behavior unchanged and
  makes the existing test-server page type-check with the React JSX style contract.

  No functional behavior changed in this component.

apps/test-server/index.tsx
walton007: Yes, for the test server we should show this on the HTML page in dev mode. A boot failure
  currently leaves the app blank with only a console error, which is not useful when testing the
  WebSpatial runtime.

  I’ll change the test-server wrapper to keep the `console.error`, but also render a visible boot
  error panel with the `WebSpatialBootError` message and cause. For SDK consumers, the intended
  action is still to handle `SpatialBoot.onError` in their app shell and show their own retry/error
  UI.

docs/migration/lazy-load-spatial-runtime.md
walton007: Agreed. I’ll remove the `@webspatial/vite-plugin` migration section from this public-facing
  guide.

  The migration doc should describe the v1 -> v2 upgrade path using the current public API shape:
  choose the default or eager React SDK entry, add `SpatialBoot` / `bootSpatial` where needed, and
  replace removed import paths/exports. Keeping plugin-specific instructions here would reintroduce
  a concept that the public docs intentionally removed.

docs/migration/lazy-load-spatial-runtime.md
walton007: That makes sense. I’ll trim this section and rewrite it around concrete upgrade actions for
  existing apps.

  Instead of explaining implementation/background details, this guide should focus on what
  developers need to change: which import root to use, where to place `SpatialBoot` /
  `bootSpatial`, how to handle SSR/eager usage, and how to safely use imperative refs after
  readiness.

docs/migration/lazy-load-spatial-runtime.md
walton007: This is a real caveat with late boot. If `<Model />` mounts before `bootSpatial()` resolves, the
  forwarded ref can initially point at the fallback `<model>` element, so `modelRef.current.ready`
  is not a reliable signal for spatial `ModelRef` APIs like `entityTransform`.

  I’ll update the migration guidance to make that explicit: do not use the fallback ref as a boot/
  readiness signal. For imperative `ModelRef` APIs, mount the spatial subtree under
  `<SpatialBoot>`, or gate/remount the `<Model />` itself after `useSpatialReady()` is true. In
  those patterns, `modelRef.current` points at the real spatial implementation before callers use
  `ready`, `entityTransform`, or other spatial-only APIs.

packages/react/src/facades/Model.tsx
walton007: Removed. The parity/spec details are already covered by tests and the migration/spec docs, so
  keeping that long implementation note in the facade makes the component harder to read without
  adding much local value.

packages/react/src/facades/Model.tsx
walton007: Yes, that can happen if the app mounts <Model /> before bootSpatial() resolves and then reads
modelRef.current.ready from a one-shot effect. The initial ref can point at the fallback
<model>, and React will later replace that node with the real spatial implementation without
re-running an effect that captured the old ref.

I’m documenting this as a late-boot integration caveat rather than trying to preserve the
fallback ref across the swap. The safe patterns are to mount the spatial subtree under
<SpatialBoot>, or gate/remount <Model /> after useSpatialReady() is true before using
ready, entityTransform, or other spatial ModelRef APIs.

packages/react/src/facades/Model.tsx
walton007: I think keeping `renderModelFallback` is still useful here because it keeps the spatial-only prop
  stripping isolated from the main facade branch.

  The part I’ll change is the non-standard fallback creation. Instead of:

  ```tsx
  return createElement('model', { ...modelProps, ref })

  I’ll add the JSX intrinsic typing needed for <model> and return normal JSX:

  return <model ref={ref} {...modelProps} />

  That keeps the component structure the same while removing the createElement('model', ...)
  pattern that is hard for developers and AI tools to read.

packages/react/src/facades/Model.tsx
walton007: Done. The fallback now returns normal JSX:

  ```tsx
  return <model ref={ref} {...modelProps} />

  I kept renderModelFallback as a helper because it keeps the spatial-only prop stripping isolated
  from the main facade branch, but removed the createElement('model', ...) workaround by adding the
  local JSX intrinsic typing needed for <model>.

packages/react/src/runtime/bridge.ts
walton007: Agreed. I’ll add an internal `requireSpatialImpl()` helper next to `getSpatialImpl()`.

  `getSpatialImpl()` will keep the nullable return for callers that need to inspect readiness,
  while `requireSpatialImpl()` will return `SpatialImplementation` and throw a clear error if it is
  called before the bridge is ready. The facade paths that only run after `useSpatialReady()` /
  `<SpatialBoot>` can use `requireSpatialImpl()` instead of `getSpatialImpl()!`.

packages/visionOS/web-spatial/manifest.swift
walton007: Reverted. This was a local manifest/debug URL change and should not be part of this PR.

docs/migration/lazy-load-spatial-runtime.md
khufu-dev: Sorry my original comment wasn't correct. It took AI couple of prompts to understand that `<SpatialBoot>` blocks the rendering of it's children and that this problem should not affect users.

packages/react/src/facades/Model.tsx
khufu-dev: Makes sense. We have to make sure developers use `<SpatialBoot>` correctly and don't document other confusing behaviours that break their code in WebSpatial mode.

packages/react/src/facades/Model.tsx
khufu-dev: Can it be a normal JSX component? In React 19 ref is a normal property.
```suggestion
function ModelFallback(
```

khufu-dev (approved):
Github cannot show me the changes since my last code review since the pull request is quite big. But I believe the changes are good.

packages/react/src/facades/Model.tsx
walton007: I think keeping this as a small helper is clearer while we still support React 18.

  The main readability issue was the non-standard `createElement('model', ...)` call. That is now
  replaced with normal JSX:

  ```tsx
  return <model ref={ref} {...modelProps} />

  Turning this into a separate ModelFallback component would still need forwardRef for React 18
  compatibility, so it would add more React wrapper code without making the fallback path easier to
  read. I’d prefer to keep renderModelFallback(props, ref) as a focused helper for stripping
  spatial-only props and returning the normal JSX fallback.

PR #1235, opened 2026-05-26
fix(react): handle missing texture ids for UnlitMaterial without breaking render

parthpatelsj:
### Motivation
- Creating an `UnlitMaterial` with a `textureId` that was never declared caused `ResourceRegistry.get()` to return a deferred promise and blocked material creation, which could make entities not render.
- The native/AVP expectation is that a missing texture should yield a tint-only material (no texture) rather than blocking render.
- The web/react layer should detect undeclared texture ids and fall back to no-texture while still allowing the texture to be bound later.

### Description
- Added `ResourceRegistry.has(id): boolean` to detect whether a resource id is known without creating a deferred `get()` placeholder.
- Updated `UnlitMaterial` initialization to check `resourceRegistry.has(options.textureId)` and use `textureId = ''` fallback when the id is undeclared so the material is created immediately.
- Updated `UnlitMaterial` dynamic property update path to apply the same undeclared-id fallback when `textureId` changes.
- Kept existing subscription/notify flow so the material can bind the texture later when the `textureId` is added.

### Testing
- Ran `pnpm -C packages/react build` which completed successfully.
- The TypeScript build/DTS generation (`tsup`) completed with no errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a162277fc188323973adf9ac7df8539)

### DIFF
--- a/.changeset/bright-pandas-bloom.md
+++ b/.changeset/bright-pandas-bloom.md
@@ -1,0 +1,5 @@
+---
+'@webspatial/react-sdk': patch
+---
+
+Fix `<UnlitMaterial>` so providing an undeclared `textureId` no longer blocks material creation. It now falls back to color-only rendering (empty native `textureId`) and continues to support binding once the texture resource is later declared.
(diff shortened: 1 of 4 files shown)


khufu-dev:
@codex review please

khufu-dev (approved):
Link the issue that this fixes in the description.

PR #1240, opened 2026-05-27
fix(visionos): sync nav button state from webview

JackHFK:
## Summary
Fix visionOS nav button state sync for SPA navigation.

## Problem
The nav UI previously refreshed `canGoBack` and `canGoForward` from URL/load-related callbacks, and also read them immediately after `goBack()` / `goForward()`.

For test-server sidebar navigation, most route changes are SPA hash-route updates. In this flow, the URL can change before `WKWebView.canGoBack` and `WKWebView.canGoForward` settle, which makes the nav button state lag one step behind the actual back-forward stack.

## Fix
- Observe `WKWebView.canGoBack` directly
- Observe `WKWebView.canGoForward` directly
- Emit a dedicated navigation-state update event from the webview layer
- Update `SpatialNavView` only from that navigation-state event
- Keep URL updates separate for address changes
- Align state-listener removal signatures with the registered callback types

## Files
- `packages/visionOS/web-spatial/view/SpatialNavView.swift`
- `packages/visionOS/web-spatial/webview/SpatialWebController.swift`
- `packages/visionOS/web-spatial/webview/SpatialWebViewModel.swift`

## Validation
- Verified Swift diagnostics for the touched files
- Branch is isolated from earlier non-navigation fixes and can be reviewed independently against `main`

image
image
image

### DIFF
--- a/packages/visionOS/web-spatial/view/SpatialNavView.swift
+++ b/packages/visionOS/web-spatial/view/SpatialNavView.swift
@@ -76,12 +76,10 @@ struct SpatialNavView: View {
     @State private var showEnter: Double = 1
     @State private var showNav: Double = 0
     @State private var showUrl: Double = 0
+    @State private var canGoBack = false
+    @State private var canGoForward = false
+    @State private var navigationStateListener: ((SpatialWebViewState) -> Void)?
     @Namespace var hoverNamespace
-
-    func checkButtonState() {
-        canGoBack = model!.getController().webview!.canGoBack
-        canGoForward = model!.getController().webview!.canGoForward
-    }
 
     func goBack() {
         spatialScene.resetForNavigation()
@@ -107,9 +105,35 @@ struct SpatialNavView: View {
         return model?.getController().webview?.url
     }
 
-    @State var canGoBack: Bool = false
+    func checkButtonState() {
+        canGoBack = model?.getController().webview?.canGoBack ?? false
+        canGoForward = model?.getController().webview?.canGoForward ?? false
+    }
 
-    @State var canGoForward: Bool = false
+    func registerNavigationStateListener() {
+        guard navigationStateListener == nil else { return }
+
+        let listener: (SpatialWebViewState) -> Void = { state in
+            switch state {
+            case .didUpdateNavigationState:
+                DispatchQueue.main.async {
+                    self.checkButtonState()
+                }
+            default:
+                break
+            }
+        }
+
+        navigationStateListener = listener
+        model?.addStateListener(listener)
+    }
+
+    func unregisterNavigationStateListener() {
+        guard let navigationStateListener else { return }
+
+        model?.removeStateListener(navigationStateListener)
+        self.navigationStateListener = nil
+    }
 
     var navHoverGroup: HoverEffectGroup {
         HoverEffectGroup(hoverNamespace)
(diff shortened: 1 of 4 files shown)


khufu-dev:
@codex review

packages/visionOS/web-spatial/view/SpatialNavView.swift
```diff
@@ -107,9 +105,35 @@ struct SpatialNavView: View {
...
+
+        navigationStateListener = listener
+        model?.addStateListener(listener)
+    }
+
+    func unregisterNavigationStateListener() {
+        guard let navigationStateListener else { return }
 
-    @State var canGoForward: Bool = false
+        model?.removeStateListener(navigationStateListener)
```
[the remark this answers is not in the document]
walton007: @codex try to fix it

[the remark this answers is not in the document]
JackHFK: Fixed in `ca311ac9` by making generic state-listener registration return a `UUID` token and removing by that token instead of comparing closure identity. `SpatialNavView` now stores the listener token and clears the exact registered listener on `onDisappear`, so repeated ornament teardown/recreate cycles should no longer accumulate stale nav-state listeners.

packages/visionOS/web-spatial/view/SpatialNavView.swift
```diff
@@ -107,9 +105,34 @@ struct SpatialNavView: View {
...
+        let listener: (SpatialWebViewState) -> Void = { state in
+            switch state {
+            case .didUpdateNavigationState:
+                DispatchQueue.main.async {
+                    self.checkButtonState()
+                }
+            default:
+                break
+            }
+        }
```
khufu-dev, line 121: Use modern Swift
```suggestion
   let listener: (SpatialWebViewState) -> Void = { [weak self] state in
        guard let self else { return }
        if case .didUpdateNavigationState = state {
            Task { @MainActor in
                self.checkButtonState()
            }
        }
    }
```

packages/visionOS/web-spatial/webview/SpatialWebController.swift
khufu-dev: Can we make this more strictly typed?

packages/visionOS/web-spatial/webview/SpatialWebController.swift
```diff
@@ -221,12 +225,20 @@ class SpatialWebController: NSObject, WKNavigationDelegate, WKScriptMessageHandl
...
         {
             DispatchQueue.main.async {
-//                print("url change", url)
                 self.model?.url = url
+                self.webviewStateChangeInvoke?(.didUpdateURL)
+            }
+        } else if keyPath == #keyPath(WKWebView.canGoBack) ||
+            keyPath == #keyPath(WKWebView.canGoForward)
+        {
+            DispatchQueue.main.async {
```
khufu-dev: Use modern Swift actors.

khufu-dev (approved):

packages/visionOS/web-spatial/view/SpatialNavView.swift
JackHFK: Updated to use `guard case` and `Task { @MainActor }`. I did not use `[weak self]` here because `SpatialNavView` is a SwiftUI `View` struct, and weak capture only applies to class-bound references.

packages/visionOS/web-spatial/webview/SpatialWebController.swift
JackHFK: Done. Replaced the string-based `observeValue(... Any?)` KVO path with typed `NSKeyValueObservation` for `\.url`, `\.canGoBack`, and `\.canGoForward`, so the callbacks no longer need `Any` casting or keyPath string switching.

PR #1267, opened 2026-06-04
Fix Vitest Dependabot alerts

walton007:
## Summary
- Upgrade root and React SDK Vitest tooling to patched 4.1.x releases.
- Add explicit Vite 6.4.2 peers where Vitest/plugin-react need them, while preserving the React 19 fixture's Vite 8 coverage.
- Update Vitest 4 test compatibility for newable mocks and source aliases.

## Verification
- `pnpm --filter @webspatial/core-sdk test` passed: 11 files, 118 tests.
- `pnpm --filter @webspatial/react-sdk test` passed: 12 files, 117 tests.
- `pnpm run check:unused` passed.
- `pnpm audit --audit-level critical` passed: no known vulnerabilities found.
- `pnpm test` is blocked by existing `apps/test-server` typecheck errors in `src/pages/jsapi-refresh-validation/index.tsx` using `Array.prototype.at` with the current TS lib target (lines 454 and 582). Package tests and workflow check passed within that run.

### DIFF
--- a/apps/test-server/package.json
+++ b/apps/test-server/package.json
@@ -68,7 +68,8 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-react-refresh": "^0.4.6",
     "postcss-modules": "^6.0.0",
-    "typescript": "^5.2.2"
+    "typescript": "^5.2.2",
+    "vite": "6.4.2"
   },
   "husky": {
     "hooks": {
--- a/packages/react/vitest.config.ts
+++ b/packages/react/vitest.config.ts
@@ -1,10 +1,22 @@
 import { defineConfig } from 'vitest/config'
 import react from '@vitejs/plugin-react'
+import { fileURLToPath, URL } from 'node:url'
 
 export default defineConfig({
   plugins: [react()],
+  resolve: {
+    alias: {
+      '@webspatial/core-sdk': fileURLToPath(
+        new URL('../core/src/index.ts', import.meta.url),
+      ),
+      '@webspatial/react-sdk': fileURLToPath(
+        new URL('./src/index.ts', import.meta.url),
+      ),
+    },
+  },
   test: {
     environment: 'jsdom',
+    exclude: ['dist/**', 'node_modules/**'],
     setupFiles: ['./vitest.setup.ts'],
   },
 })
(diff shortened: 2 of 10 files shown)


apps/test-server/package.json
```diff
@@ -68,7 +68,8 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-react-refresh": "^0.4.6",
     "postcss-modules": "^6.0.0",
-    "typescript": "^5.2.2"
+    "typescript": "^5.2.2",
+    "vite": "6.4.2"
```
khufu-dev, line 72: Upgrade to Vite 8 (latest) for testing. Most new developers would use latest version.

packages/react/vitest.config.ts
```diff
@@ -1,10 +1,22 @@
...
   plugins: [react()],
+  resolve: {
+    alias: {
+      '@webspatial/core-sdk': fileURLToPath(
+        new URL('../core/src/index.ts', import.meta.url),
+      ),
+      '@webspatial/react-sdk': fileURLToPath(
+        new URL('./src/index.ts', import.meta.url),
+      ),
+    },
```
khufu-dev, line 15: AI puts this in but I don't think we need it because we're using pnpm.

khufu-dev (approved):

apps/test-server/package.json
walton007: Keeping this at `6.4.2` is intentional for this PR. Vitest 4.1.x declares `vite >=6.4.2 <7.0.0`, and `@vitejs/plugin-react@4.4.1` is also in that compatibility band. When this floated to Vite 8 locally, pnpm resolved unrelated plugin/Vitest peers through the React 19 fixture's Vite 8 install and produced peer warnings.

The latest-Vite coverage still exists in `tests/react19-compat` with `vite 8.0.10`; this PR is scoped to clearing the Vitest security alert without expanding the test-server toolchain upgrade.

packages/react/vitest.config.ts
walton007: This is still needed with pnpm. pnpm links the workspace package, but the package `exports` for `@webspatial/core-sdk` and the self-reference `@webspatial/react-sdk` still point at `dist`. Under Vitest 4, that made the React package tests resolve package entries instead of source, and stale `dist` test files were also being picked up.

The aliases force these package tests to exercise source files, and `exclude: ['dist/**', 'node_modules/**']` prevents generated output from being collected as tests. Without this, the React package tests fail unless the dist state happens to line up with the source.

PR #1275, opened 2026-06-08
spatialweb-event support multi-receivers

mzvast:
## Summary

This PR extracts the previously reviewed `SpatialWebEvent` multi-receiver change from #1136 into an independent PR targeting `main`.

The original change was merged into a non-main branch as part of #1136, but that branch has since removed this change. This PR reintroduces the same behavior as a standalone change so it can be reviewed and merged independently.

This branch also includes a follow-up cleanup to keep the new receiver registry encapsulated: `SpatialWebEvent.eventReceiver` is now private, and tests assert observable behavior instead of reading the internal storage directly.

## Background

`SpatialWebEvent` is the shared bridge used to route native-originated event payloads back into JavaScript. Some SDK subsystems register listeners under shared event ids, such as the `"window"` id used by physical metrics updates.

Previously, each event id could only hold one receiver callback. This meant:

* Later registrations overwrote earlier receivers for the same id.
* Removing one subscriber could accidentally remove the shared receiver and break other subscribers.
* Multiple consumers of the same native event stream could interfere with each other.

This was originally addressed in #1136 and reviewed there. Since that PR did not land in `main`, this PR separates the change and brings it back as an independent patch.

## Changes

* Change `SpatialWebEvent` receiver storage from one callback per id to multiple callbacks per id.
* Dispatch each incoming event payload to every receiver currently registered under the event id.
* Support targeted cleanup via `removeEventReceiver(id, callback)`, which removes only the specified receiver.
* Preserve the existing full-cleanup behavior of `removeEventReceiver(id)`, which removes all receivers for that id.
* Remove the id entry after the last receiver is removed to avoid retaining empty receiver sets.
* Keep the public API compatible for existing single-receiver call sites.
* Make the internal receiver registry private so callers and tests do not depend on the backing data structure.
* Replace direct internal-state assertions with behavior-driven coverage for cleanup and unsubscribe semantics.
* Add OpenSpec proposal, design, and requirement docs for the multi-receiver behavior.
* Add/update test coverage for multi-subscriber behavior and cleanup semantics.

## Compatibility

This should be backward compatible for existing callers:

* Existing `addEventReceiver(id, callback)` usage continues to work.
* Existing `removeEventReceiver(id)` usage still removes all receivers for the id.
* Callers that share an id can now use `removeEventReceiver(id, callback)` to unsubscribe without affecting other receivers.

## Testing

This PR includes behavior coverage for:

* Multiple receivers registered under the same `SpatialWebEvent` id.
* Fan-out dispatch to all active receivers.
* Removing one receiver without affecting the others.
* Stopping delivery once the last receiver is removed.
* `SpatializedElement.destroy()` no longer receiving events after cleanup.
* Pico OS request correlation cleanup paths on success and timeout.
* Existing single-receiver cleanup paths.

Validation run for this branch:

* `pnpm --filter @webspatial/core-sdk test`
* `pnpm exec tsc -p apps/test-server/tsconfig.json`
* `npm run test`

## Related PR

* Based on the previously reviewed change in #1136.

### DIFF
--- a/.changeset/fifty-geckos-hear.md
+++ b/.changeset/fifty-geckos-hear.md
@@ -1,0 +1,5 @@
+---
+'@webspatial/core-sdk': patch
+---
+
+`SpatialWebEvent` now supports multiple receivers per id
--- a/openspec/changes/spatial-web-event-multi-receivers/.openspec.yaml
+++ b/openspec/changes/spatial-web-event-multi-receivers/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-08
--- a/openspec/changes/spatial-web-event-multi-receivers/design.md
+++ b/openspec/changes/spatial-web-event-multi-receivers/design.md
@@ -1,0 +1,99 @@
+## Context
+
+`SpatialWebEvent` is the SDK's shared bridge for routing native-originated event payloads back into JavaScript. Multiple parts of the SDK register listeners under shared ids such as scene objects, window-level metric listeners, and one-shot platform adapter callbacks. This branch changes the internal storage from one callback per id to many callbacks per id, so the OpenSpec design needs to define registration, dispatch, and cleanup semantics explicitly.
+
+## Goals / Non-Goals
+
+**Goals:**
+- Allow one event id to fan out a payload to multiple registered receivers.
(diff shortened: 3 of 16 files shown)


mzvast:
closed due to @walton007 think the event system should not support multiple listener.

PR #1231, opened 2026-05-24
feat: Radix floating menu support for main page and SpatialDiv

walton007:
## Summary

This PR adds the Radix floating-menu integration path for WebSpatial SpatialDivs and extends it to plugin-hosted / dual-root menus.

It now covers five demo scenarios on `#/dropdown-menu-spatial`:

| Scenario | Pattern | Result |
| --- | --- | --- |
| 1 | Main-page `DropdownMenu.Content asChild` + inner `div enable-xr` | Menu rises as a root spatial surface |
| 2 | Menu inside a parent SpatialDiv + `useSpatialPortalContainer()` | Flat Radix menu stays in the parent spatial window |
| 3 | Nested `div enable-xr data-xr-overlay` inside Radix `Content asChild` | Menu rises as a child spatial surface and escapes parent bounds |
| 4 | Flat-page plugin host + `SpatialOverlay` | Plugin items bridge into measurement and visible portal targets |
| 5 | Plugin host inside parent SpatialDiv + `SpatialOverlay` + `data-xr-overlay` | Plugin menu rises as a child spatial surface inside the parent SpatialDiv |

Related issues:

- Addresses #1055
- Addresses #1286

> AVP iwdp tooling is generic simulator/debug tooling and remains separate in #1283.

## SDK highlights

- Add `SpatialWindowContext` and `useSpatialPortalContainer()` so Radix/Floating UI content can portal into the nearest spatial window instead of the main page document.
- Add declarative child-surface overlay mode via `data-xr-overlay` on nested `enable-xr` content.
- Add hidden placeholder dual-render for child overlays so Radix `asChild` ref/props and measurement land in the parent spatial window while visible content renders in the child spatial webview.
- Attach floating overlays to the parent spatial element and use raw placeholder rects for parent-window coordinates.
- Decouple overlay native visibility from hidden measurement placeholders.
- Add `SpatialOverlay`, `useSpatialOverlay()`, and `SpatialOverlayPortalOption` for plugin-hosted / dual-root menus.
- Add internal `SpatialOverlayRenderTargetContext` so `SpatialOverlay` can distinguish measurement placeholder renders from visible portal renders, including nested Scenario 5.
- Make `portalMenuOption(content)` self-subscribing so plugin content rendered before targets mount appears once the measurement/portal targets are available.

## Developer API notes

Scenario 2:

```tsx
const container = useSpatialPortalContainer()

<DropdownMenu.Portal container={container}>
[...]

PR #1276, opened 2026-06-08
fix(visionos): recreate scene when WindowGroup data is stale

walton007:
## Summary

- Add `getSceneOrCreate()` in `WebSpatialApp` so window/volume `WindowGroup` entries recreate a spatial scene when the stored scene ID is missing or stale, instead of force-unwrapping nil.
- Read volume resize limits from `spatialScene.sceneConfig` so recreated scenes use the active scene configuration.

## Context

This is a visionOS shell **defensive** fix, independent of Radix / SpatialDiv portal work (#1231).

Today `WebSpatialApp` does:

```swift
let spatialScene = SpatialApp.Instance.getScene(windowData)
SpatialSceneView(spatialScene: spatialScene!)
```

If SwiftUI `WindowGroup(for: String.self)` re-renders with a scene ID that is no longer in `SpatialApp.scenes` (for example after `SceneHandlerUIView.onDisappear` calls `spatialScene.destroy()`), the force-unwrap crashes the app.

This change removes the crash by recreating a scene when lookup fails. It was originally bundled into #1231 and split out here to keep that PR focused on `useSpatialPortalContainer()`.

## Repro status

**No stable manual repro is documented in the repo.** Candidate paths below are plausible from code review but are **not guaranteed** to crash on `main`:

**Path A — `#/scene` → `window.open` → close child window**

Normal close order is usually `dismissWindow` while the scene still exists, then `destroy()` on disappear. Manual testing has **not** reliably reproduced a crash on `main`.

**Path B — Xcode Stop/Run without resetting Simulator**

Process memory is cleared while WindowGroup state may restore a stale scene ID. Worth trying, but also not confirmed as a stable repro.

**Not related:** SPA back/forward/reload via `resetForNavigation()` destroys page spatial objects, not entries in `SpatialApp.scenes`.

## Verify fix

- [ ] Build `web-spatial` for visionOS Simulator
- [ ] Smoke-test cold start, `#/scene`, and child `window.open` / close on this branch
- [ ] Optional: try Path A/B on `main` vs this branch; a successful fix may only show as Console log `Missing scene for WindowGroup data. Recreating scene.` rather than a visible crash
- [ ] Confirm normal navigation still works after merge

## Review note

Treat this as hardening against a rare WindowGroup/scene lifecycle race, not as a fix for a confirmed always-repro user bug. Merging is reasonable if the nil-safe behavior and smoke tests look good, even without a deterministic repro.

### DIFF
--- a/.changeset/visionos-windowgroup-scene-recreate.md
+++ b/.changeset/visionos-windowgroup-scene-recreate.md
@@ -1,0 +1,7 @@
+---
+'@webspatial/platform-visionos': patch
+---
+
+Recreate the spatial scene when visionOS `WindowGroup` window data references a missing scene ID, instead of force-unwrapping nil. Read volume resize limits from the active scene config after recreation.
+
+---
--- a/packages/visionOS/web-spatial/WebSpatialApp.swift
+++ b/packages/visionOS/web-spatial/WebSpatialApp.swift
@@ -40,13 +40,31 @@ struct WebSpatialApp: App {
         )
     }
 
+    func getSceneOrCreate(_ sceneID: String?, _ style: SpatialScene.WindowStyle) -> SpatialScene {
+        if let sceneID, let spatialScene = SpatialApp.Instance.getScene(sceneID) {
+            return spatialScene
+        }
+
+        logger.debug("Missing scene for WindowGroup data. Recreating scene.")
+        return SpatialApp.Instance.createScene(
+            startURL,
+            style,
+            .visible,
+            app.getSceneOptions()
+        )
+    }
+
+    func getSceneOrCreate(_ sceneID: String, _ style: SpatialScene.WindowStyle) -> SpatialScene {
+        return getSceneOrCreate(Optional(sceneID), style)
+    }
+
     var body: some Scene {
         WindowGroup(
             id: SpatialScene.WindowStyle.window.rawValue,
             for: String.self
         ) { $windowData in
-            let spatialScene = SpatialApp.Instance.getScene(windowData)
-            SpatialSceneView(spatialScene: spatialScene!)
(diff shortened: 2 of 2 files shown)


mzvast (changes requested):
Thanks for splitting this out. The nil-safe fallback is a good direction, and reading the volume resize range from `spatialScene.sceneConfig` looks right because it follows the actual scene being rendered after recreation.

I think this needs one more fix before merge: when `windowData` contains a stale scene id, `getSceneOrCreate` creates a replacement scene but does not update the `WindowGroup` value. On a later SwiftUI reevaluation, the same stale id will miss again and create another `SpatialScene`. Since `SpatialSceneView` stores the initial scene in `@State`, those later replacement scenes may not be adopted by the current view and can remain registered in `SpatialApp.scenes`, potentially leaking web views/scenes and making `firstActiveScene` selection ambiguous.

Could we write the recreated `scene.id` back to the WindowGroup binding, or memoize stale id -> recreated scene id, so the fallback is stable across reevaluations?

One extra note: fallback recreation uses `startURL` + current `app.getSceneOptions()`, so it may not restore the original child window URL/config. That is probably acceptable for a defensive crash hardening fix, but it would be good to make that limitation explicit in the PR description or code comment.

mzvast (approved):
Rechecked the scope: if the expectation of this PR is defensive hardening to avoid the stale `WindowGroup` scene id crash, the current change is acceptable.

The stale id -> recreated scene id memoization / binding write-back concern is still worth tracking, but I do not think it needs to block this PR unless we want stronger guarantees around repeated SwiftUI body reevaluation and scene restoration stability.

Approving this as a crash-hardening fix.

PR #1281, opened 2026-06-09
fix: change entity naming standard

parthpatelsj:
Hyphenated WebSpatial entity name values cause cross-platform problems — Reality Composer Pro normalizes unsupported characters to underscores, which can break serialization, USD export, and scene loading. This PR adds compile-time guardrails so authors catch bad names early.

- Reject literal hyphenated name values in TypeScript (core APIs and React entity components like BoxEntity, ModelEntity, etc.)
- Show a clear compiler error for JSX literals, e.g. "invalid-box" → Entity names must not include hyphens… Use camelCase or underscores instead
- Document the naming policy in AGENTS.md, docs/dynamic-3d-api-prd.md, and docs/webspatial-quirks.md
Add OpenSpec skill guardrails so agents don’t generate hyphenated entity names in examples
Breaking change (major bump): rename hyphenated names to camelCase or underscores — e.g. robot-arm → robotArm or robot_arm.

Scope note: enforcement is compile-time only on name (not id). Dynamic string values still type-check; the policy is enforced where TypeScript can infer the literal.

### DIFF
--- a/.changeset/strict-entity-names.md
+++ b/.changeset/strict-entity-names.md
@@ -5,6 +5,6 @@
 
 Disallow hyphens in WebSpatial entity names.
 
-Entity `name` values must now use USD-safe identifiers without hyphens. Literal hyphenated names are rejected by TypeScript where the name can be inferred, and dynamic string values are validated at runtime before native entity creation.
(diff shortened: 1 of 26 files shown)


walton007:
@codex please review the PR

parthpatelsj:
Not to be fixed, as this is throwing a warning - and not an exception (which crashes the app)

PR #1290, opened 2026-06-16
fix(core): persist html material bracket assignment

### LINKED ISSUE #1289: CSS custom property bracket assignment can emit stale computed material after html style mutation
### Summary

In `spatial-window-polyfill`, assigning the scene background material through bracket syntax on `document.documentElement.style` can emit the selected material first, but a following unrelated html style mutation may cause the SDK to emit `material: "none"`.

Example:

```ts
document.documentElement.style["--xr-background-material"] = "translucent";
document.documentElement.style["border-radius"] = "80px";
```

In one observed case, the stale computed value was empty, so the SDK fell back to `"none"`:

```txt
UpdateSpatialSceneProperties, {"material":"translucent"}
UpdateSpatialSceneProperties, {"cornerRadius":{...}}
UpdateSpatialSceneProperties, {"material":"none"}
```

The last event incorrectly resets the scene background material back to `none`.

### Root Cause

`spatial-window-polyfill.ts` proxies `document.documentElement.style`.

For direct style assignment, the current `set` trap first writes through `Reflect.set`:

```ts
set: function (target, key, value) {
  const ret = Reflect.set(target, key, value)

  if (key === SpatialGlobalCustomVars.backgroundMaterial) {
    setCurrentWindowStyle(value)
  }

  // border-radius / opacity handling...

  return ret
}
```

This means bracket assignment to the CSS custom property:

```ts
document.documentElement.style["--xr-background-material"] = "translucent";
```

causes the SDK to immediately emit:

```ts
updateSpatialProperties({ material: "translucent" })
```

However, `Reflect.set(target, "--xr-background-material", value)` does not reliably update the actual CSS custom property in `CSSStyleDeclaration`. CSS custom properties are supposed to be written through:

```ts
style.setProperty("--xr-background-material", value)
```

After that, writing a normal inline style such as:

```ts
document.documentElement.style["border-radius"] = "80px";
```

does update the real style attribute and triggers the polyfill's `MutationObserver`.

The observer then calls `checkCSSProperties()`, which reads:

```ts
getComputedStyle(document.documentElement)
  .getPropertyValue("--xr-background-material")
```

Since the bracket assignment does not reliably write the CSS custom property into CSSOM, a later html style mutation causes the MutationObserver to recompute the material from `getComputedStyle(document.documentElement)`.

At that point, the SDK may read the previous computed material value rather than the value assigned through bracket syntax. If no computed value exists, the current fallback maps it to `"none"`:

```ts
setCurrentWindowStyle(backgroundMaterial || "none")
```

So `material: "none"` is one concrete symptom of the broader issue: the SDK can emit a stale computed material value after another html style mutation.

### Why this is first-trigger sensitive

The first `border-radius` update changes the actual inline style attribute, so the `MutationObserver` recomputes all tracked html-level properties.

If the same `border-radius` value is written again later, it may not produce the same effective style mutation. As a result, the stale `material: "none"` emission is most visible on the first trigger.

### Expected Behavior

Both of these should keep the SDK's material state and computed style state consistent:

```ts
document.documentElement.style.setProperty(
  "--xr-background-material",
  "translucent"
);
```

```ts
document.documentElement.style["--xr-background-material"] = "translucent";
```

After bracket assignment, subsequent html style mutations should not cause the SDK to emit a material value derived from stale computed style. The computed style should reflect the assigned custom property, or the SDK should otherwise avoid treating stale computed values as a new material update.

### Proposed Fix

In the `style` proxy `set` trap, handle `--xr-background-material` as a CSS custom property and write it through `CSSStyleDeclaration.setProperty`.

The fix should preserve the existing side-effect flow and avoid returning early from the custom-property branch.

Suggested shape:

```ts
set: function (target, key, value) {
  let ret: boolean

  if (key === SpatialGlobalCustomVars.backgroundMaterial) {
    // Write through the raw CSSStyleDeclaration target, not through the proxied
    // document.documentElement.style, so this does not re-enter the proxy traps.
    target.setProperty(SpatialGlobalCustomVars.backgroundMaterial, String(value))
    ret = true
  } else {
    ret = Reflect.set(target, key, value)
  }

  if (key === SpatialGlobalCustomVars.backgroundMaterial) {
    setCurrentWindowStyle(String(value))
  }

  if (
    key === "border-radius" ||
    key === "borderRadius" ||
    key === "border-top-left-radius" ||
    key === "borderTopLeftRadius" ||
    key === "border-top-right-radius" ||
    key === "borderTopRightRadius" ||
    key === "border-bottom-left-radius" ||
    key === "borderBottomLeftRadius" ||
    key === "border-bottom-right-radius" ||
    key === "borderBottomRightRadius"
  ) {
    checkCornerRadius()
  }

  if (key === "opacity") {
    checkOpacity()
  }

  return ret
}
```

### Why this should not recurse

Inside the `set` trap, `target` is the raw `CSSStyleDeclaration` captured before installing the proxy:

```ts
const rawDocumentStyle = document.documentElement.style
```

Calling:

```ts
target.setProperty("--xr-background-material", String(value))
```

uses the raw target method directly. It does not go through the proxied `document.documentElement.style`, so it does not re-enter the proxy `get` or `set` traps.

It may still trigger a normal `style` attribute mutation, which is expected. But after the fix, computed style can read the correct material value. Since `setCurrentWindowStyle(String(value))` has already updated the cached material, the observer's later recomputation should see the same value and should not emit another material update.

The flow becomes:

```txt
style["--xr-background-material"] = "translucent"
  -> proxy set trap
  -> raw target.setProperty("--xr-background-material", "translucent")
  -> setCurrentWindowStyle("translucent")
  -> emit material:"translucent"
  -> style mutation observer fires
  -> checkHtmlBackgroundMaterial()
  -> computed style reads "translucent"
  -> setCurrentWindowStyle("translucent")
  -> no extra emit because the cached value is unchanged
```

### Additional Notes

The application can avoid this issue by using the standard CSSOM API directly:

```ts
document.documentElement.style.setProperty(
  "--xr-background-material",
  "translucent"
);
```

However, the SDK currently treats bracket assignment as a supported material update path by intercepting `style["--xr-background-material"]`. Therefore the proxy should make bracket assignment semantically equivalent to `setProperty`.

JackHFK:
## Summary

Fix html-level background material updates when `--xr-background-material` is assigned through bracket syntax on `document.documentElement.style`.

Before this change, bracket assignment could update the SDK's internal material state and emit a material update, but fail to persist the CSS custom property into CSSOM. A later html style mutation could then cause the polyfill to recompute material from `getComputedStyle(document.documentElement)` and emit a stale computed material value.

In the reported case, the stale computed value was empty, so the fallback emitted `material: "none"`. The broader issue is that the recomputed material could come from stale computed style rather than the value assigned through bracket syntax.

Fixes #1289

## Root Cause

`spatial-window-polyfill` intercepts `document.documentElement.style` writes through a Proxy.

For bracket assignment to CSS custom properties, the previous implementation used `Reflect.set(...)`. That path could trigger the SDK's material update logic, but it did not reliably persist the CSS custom property into CSSOM.

After a later html style mutation, such as `border-radius`, the MutationObserver recomputes tracked html-level properties from:

```ts
getComputedStyle(document.documentElement)
```

Because the bracket-assigned custom property was not reliably present in computed style, the SDK could read and emit a stale computed material value. If the computed value was empty, the existing fallback converted it to `"none"`.

## Changes

- Persist `--xr-background-material` bracket assignments through the raw `CSSStyleDeclaration.setProperty(...)` path.
- Preserve the existing Proxy `set` trap flow, including border-radius and opacity handling.
- Keep the material cache and computed style state aligned so later html style mutations do not emit stale material values.
- Add a regression test covering:
  - bracket assignment for `--xr-background-material`
  - CSSOM/computed-style persistence
  - later html style mutation not emitting stale computed material
- Add a patch changeset for `@webspatial/core-sdk`.

## Verification

- `pnpm --filter @webspatial/core-sdk test`
- `pnpm --filter @webspatial/core-sdk run lint`

### DIFF
--- a/.changeset/fix-html-material-bracket-assignment.md
+++ b/.changeset/fix-html-material-bracket-assignment.md
@@ -1,0 +1,7 @@
+---
+'@webspatial/core-sdk': patch
+---
+
+Fix html background material updates written through bracket syntax so CSS custom
+properties are stored in CSSOM before later html style mutations recompute scene
+material.
--- a/packages/core/src/coverage-boost.test.ts
+++ b/packages/core/src/coverage-boost.test.ts
@@ -484,6 +484,13 @@ describe('spatialWindowPolyfill', () => {
         .getPropertyValue('--xr-background-material')
         .trim(),
     ).toBe('translucent')
+    expect(
+      (
+        (document.documentElement.style as any)[
+          '--xr-background-material'
+        ] as string
+      ).trim(),
+    ).toBe('translucent')
 
     updateSpatialProperties.mockClear()
     document.documentElement.style.setProperty('border-radius', '80px')
--- a/packages/core/src/coverage-boost.test.ts
+++ b/packages/core/src/coverage-boost.test.ts
@@ -470,6 +470,37 @@ describe('spatialWindowPolyfill', () => {
     expect(updateSpatialProperties).toHaveBeenCalledWith({
       material: 'none',
     })
+
+    await Promise.resolve()
+    await Promise.resolve()
+    updateSpatialProperties.mockClear()
+    ;(document.documentElement.style as any)['--xr-background-material'] =
+      'translucent'
+    expect(updateSpatialProperties).toHaveBeenCalledWith({
+      material: 'translucent',
+    })
+    expect(
+      getComputedStyle(document.documentElement)
+        .getPropertyValue('--xr-background-material')
+        .trim(),
+    ).toBe('translucent')
+
+    updateSpatialProperties.mockClear()
+    document.documentElement.style.setProperty('border-radius', '80px')
+    await Promise.resolve()
+    await Promise.resolve()
+
+    expect(updateSpatialProperties).toHaveBeenCalledWith({
+      cornerRadius: {
+        topLeading: 80,
+        topTrailing: 80,
+        bottomLeading: 80,
+        bottomTrailing: 80,
+      },
+    })
+    expect(updateSpatialProperties).not.toHaveBeenCalledWith({
+      material: 'none',
+    })
   })
 })
 
--- a/packages/core/src/spatial-window-polyfill.ts
+++ b/packages/core/src/spatial-window-polyfill.ts
@@ -116,6 +116,12 @@ function hijackDocumentElementStyle() {
       return ret
     },
     get: function (target, prop: string) {
+      if (prop === SpatialGlobalCustomVars.backgroundMaterial) {
+        return target.getPropertyValue(
+          SpatialGlobalCustomVars.backgroundMaterial,
+        )
+      }
+
       if (typeof target[prop as keyof CSSStyleDeclaration] === 'function') {
         return function (this: any, ...args: any[]) {
           if (prop === 'setProperty') {
--- a/packages/core/src/spatial-window-polyfill.ts
+++ b/packages/core/src/spatial-window-polyfill.ts
@@ -78,10 +78,20 @@ function hijackDocumentElementStyle() {
   const rawDocumentStyle = document.documentElement.style
   const styleProxy = new Proxy(rawDocumentStyle, {
     set: function (target, key, value) {
-      const ret = Reflect.set(target, key, value)
+      let ret: boolean
 
       if (key === SpatialGlobalCustomVars.backgroundMaterial) {
-        setCurrentWindowStyle(value)
+        target.setProperty(
+          SpatialGlobalCustomVars.backgroundMaterial,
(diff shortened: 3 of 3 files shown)


JackHFK:
@codex review commit: [62448ba](https://github.com/webspatial/webspatial-sdk/runs/81605104823)

walton007 (approved):

khufu-dev (approved):
Amazing 🤩

PR #1297, opened 2026-06-22
[codex] fix Model ready DOM binding

### LINKED ISSUE #1296: Model event.target.ready throws inside nested enable-xr containers
## Problem

`Model` load/ref APIs can fail when a spatial Model is nested under another spatial container:

```tsx
<div enable-xr>
  <Model enable-xr onLoad={event => event.target.ready} />
</div>
```

In this structure, React renders separate standard and portal branches. The standard branch can expose the DOM/ref target, while the actual `SpatializedStatic3DElement` is created by the portal branch.

The previous `ready` getter only read the current component instance's local `promiseRef.current`. For the standard branch this can be `null`, causing `event.target.ready` to throw instead of returning a promise.

## Expected

`event.target.ready` and `ref.current.ready` should resolve against the spatialized element actually bound to the DOM target, even when the Model is nested under another `enable-xr` container.

## Repro shape

```tsx
<SpatialBoot>
  <div enable-xr>
    <Model enable-xr onLoad={event => event.target.ready} />
  </div>
</SpatialBoot>
```

## Notes

This issue is scoped to fixing the runtime bug. A separate follow-up can clean up the internal DOM-to-spatialized-element binding mechanism.

walton007:
## Summary

Fixes #1296.

Fixes `Model` load/ref APIs when a spatial Model is nested under another spatial container, for example:

```tsx
<div enable-xr>
  <Model enable-xr onLoad={event => event.target.ready} />
</div>
```

In this structure, React renders separate standard and portal branches. The standard branch can expose the DOM/ref target, while the actual `SpatializedStatic3DElement` is created by the portal branch. The previous `ready` getter only read the current component instance's `promiseRef.current`, which can be `null` for the standard branch and throw when `event.target.ready` is accessed.

This change makes Model ref APIs resolve the spatial element from the DOM-linked binding first:

```ts
dom.__spatializedElement ?? dom.__innerSpatializedElement?.()
```

and keeps the existing local promise fallback for early root `ref.current.ready` reads before DOM binding has completed.

## Changes

- Add a shared helper for resolving the `SpatializedStatic3DElement` from a Model DOM/ref proxy.
- Update Model ref APIs (`ready`, `currentSrc`, transform, playback controls, timing fields) to use the DOM-linked spatialized element.
- Make `ready` reject with a Model load failure event instead of throwing when no ready source exists.
- Add a regression test for the nested standard-branch case where the DOM target has an inner spatialized element but no local creation promise.
- Add a test-server demo at `/static-3d-model/nested-ready` for manual verification in visionOS / WebSpatial.

## Testing

- `pnpm --filter @webspatial/react-sdk test`
- `pnpm --filter @webspatial/react-sdk exec tsc -p ./tsconfig.json --pretty false`
- `pnpm --filter web-content test`
- `pnpm --filter web-content build`
- `git diff --check`

## Reviewer note

This PR intentionally does not replace the internal `__spatializedElement` / `__innerSpatializedElement` DOM bindings. A follow-up cleanup can move that binding to a WeakMap-backed internal API; this PR only fixes the runtime bug with minimal surface area.

### DIFF
--- a/packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx
+++ b/packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx
@@ -82,6 +82,20 @@ function collectSources(children: React.ReactNode): ModelSource[] {
   return sources
 }
 
+type Static3DDomProxy = SpatializedStatic3DElementRef & {
+  __spatializedElement?: SpatializedStatic3DElement
+  __innerSpatializedElement?: () => SpatializedStatic3DElement | undefined
+}
+
+function getDomSpatializedStaticElement(
+  domProxy: SpatializedStatic3DElementRef,
+): SpatializedStatic3DElement | undefined {
+  const proxy = domProxy as Static3DDomProxy
+  // Nested standard branches expose the DOM/ref while the portal branch owns
+  // the actual static 3D element, so resolve through either DOM binding.
+  return proxy.__spatializedElement ?? proxy.__innerSpatializedElement?.()
+}
+
 function SpatializedContent(props: SpatializedStatic3DContentProps) {
   const {
     src,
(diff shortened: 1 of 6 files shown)


packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx
```diff
@@ -193,81 +207,66 @@ function SpatializedStatic3DElementContainerBase(
...
         get ready(): Promise<ModelLoadEvent> {
-          return promiseRef
-            .current!.then(spatializedElement => spatializedElement.ready)
+          const spatializedElement = getDomSpatializedStaticElement(domProxy)
+          const readySource = spatializedElement
+            ? Promise.resolve(spatializedElement)
+            : promiseRef.current
+          if (!readySource) {
+            return Promise.reject(createLoadFailureEvent(() => domProxy))
+          }
```
khufu-dev, line 220: I don't think `!readySource` check is needed because `promiseRef.current` should be available or it's a bug I need to fix.

```suggestion
            : promiseRef.current!
```

khufu-dev (approved):
I've been planning to refactor `SpatializedStatic3DElementContainerBase` to use spatializedElement directly from `createSpatializedElement()` instead of using it from the domProxy. But this is a good interim fix.

packages/react/src/spatialized-container/SpatializedStatic3DElementContainer.tsx
walton007: Thanks! That refactor sounds like the right long-term direction.

I’d like to land this PR first as a minimal fix for the current bug, and then we can continue with the broader refactor afterward.

PR #1311, opened 2026-07-03
[codex] Document portal CSS quirks

walton007:
## Summary

- Add a `UI library portals and scoped CSS` section to `docs/webspatial-quirks.md`.
- Explain why overlay styles that depend on host-page ancestors can fail inside WebSpatial child webviews.
- Recommend attaching essential styles to the visible portal node via stable classes, data attributes, UI-library content class props, or CSS variables.

## Validation

- `/Users/bytedance/github/webspatial-sdk/node_modules/.bin/prettier --check docs/webspatial-quirks.md`
- `node ./tools/scripts/check-filesize.js docs/webspatial-quirks.md`
- `node ./tools/scripts/check-valid-characters.js docs/webspatial-quirks.md`
- `git diff --check`

Note: this is a docs-only change. The local pre-commit hook could not run in the temporary worktree because `lint-staged` was not installed there, so the equivalent single-file checks above were run manually.

### DIFF
--- a/docs/webspatial-quirks.md
+++ b/docs/webspatial-quirks.md
@@ -43,13 +43,13 @@ A `<div enable-xr>` (SpatialDiv) is **not** a single DOM node for styling purpos
 
 ### What to use instead
 
-| Approach | Works for spatial transform |
-| --- | --- |
-| Class selector (`.panel { transform: … }`) | Yes — `className` is mirrored host → probe |
-| Inline `style={{ transform: … }}` | Yes — applied to the probe |
-| `ref.current.style.transform = '…'` | Yes — forwarded to the probe |
-| Tag selector on matching intrinsic (`h1 {}` with `<h1 enable-xr>`) | Yes (after probe tag mirror) |
-| Ancestor + tag (`.page h1 {}`) | Often **no** |
+| Approach                                                           | Works for spatial transform                |
+| ------------------------------------------------------------------ | ------------------------------------------ |
+| Class selector (`.panel { transform: … }`)                         | Yes — `className` is mirrored host → probe |
+| Inline `style={{ transform: … }}`                                  | Yes — applied to the probe                 |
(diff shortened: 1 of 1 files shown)


docs/webspatial-quirks.md
```diff
@@ -71,7 +71,52 @@ A `<div enable-xr>` (SpatialDiv) is **not** a single DOM node for styling purpos
...
+<body>
+  <div role="dialog" class="semi-modal-content">...</div>
+</body>
+```
+
+**Root cause:** The portal webview can receive the relevant `<head>` styles, but it does **not** preserve the host page's ancestor DOM tree. Any selector that depends on wrappers such as `.app-modal`, `.semi-portal`, `.semi-modal-wrap`, or `.semi-modal-small` may stop matching once the visible content is rendered in a separate portal target.
+
+**What to use instead:**
+
+- Put essential styles on the **actual visible node** via stable `className`, `contentClassName`, `modalContentClass`, `overlayClassName`, or a `data-*` attribute provided by the UI library.
```
khufu-dev, line 101: This surprises a lot of existing web developers and I'm hoping we can find a long term solution. Good that it's documented.

khufu-dev (approved):

docs/webspatial-quirks.md
walton007: Agreed. One caveat is that with the current child-document portal architecture, we cannot make host-page ancestor selectors transparently keep working: the visible portal node no longer has those ancestors, even if the stylesheet itself is synced.

So the long-term solution is probably not preserving arbitrary ancestor-dependent CSS semantics, but defining a clearer SDK styling contract for spatial portals: for example documented hooks for classes/variables on the portal-rendered node, and possibly reliable root/theme-variable propagation into the child document where that is safe. This doc is meant to capture the current mitigation until we have that contract.

PR #1312, opened 2026-07-03
chore: unified ESLint pre-commit governance (core & react)

mzvast:
## Summary

Consolidate ESLint governance at the workspace root and wire it into the
pre-commit flow, then extend that governance to `packages/core` and
`packages/react`.

**Before:** ESLint was split across the repo — `apps/test-server` used an
ESLint 8 `.eslintrc.cjs`, `tests/ci-test` used an ESLint 9 flat config, and
`packages/cli` had a lint script but no complete local setup. Root
`lint-staged` only formatted files and ran guard scripts, so unused imports
could still land in commits.

**After:** a single root ESLint 9 flat config (`eslint.config.js`) owns lint
behavior; package lint scripts delegate to `pnpm --workspace-root exec eslint`;
and root `lint-staged` runs `eslint --fix` before `prettier --write` for
governed directories. The shared rule surface explicitly enforces only
`unused-imports/no-unused-imports`; existing React Hooks and React Refresh
rules remain scoped to the directories that already used them.

### What changed

- Add root `eslint.config.js` as the single ESLint entry point.
- Remove split configs: `apps/test-server/.eslintrc.cjs` and
  `tests/ci-test/eslint.config.js`.
- Move ESLint dependencies to the root; package lint scripts delegate to the
  root installation.
- Add `eslint-plugin-unused-imports` and enable automatic unused-import removal.
- Use the latest ECMAScript parser target rather than restricting lint parsing
  to ES2020.
- Extend `lint-staged` and `pnpm lint` to `apps/test-server`, `tests/ci-test`,
  `packages/cli/src`, `packages/core/src`, and `packages/react/src`.

### Conservative `core` / `react` adoption

Both packages enforce only `unused-imports/no-unused-imports` through ESLint.

- **core:** its original TypeScript validation remains part of `lint` through
  `pnpm run typecheck && ...eslint...`; `typecheck` is also available as a
  separate command.
- **react:** React Hooks rules are intentionally not added to this package;
  `reportUnusedDisableDirectives` remains off to preserve existing disable
  comments. The only source edit removes three unused type imports in two test
  files.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Tests or tooling
- [x] Chore or maintenance

## Validation

- [ ] `pnpm test`
- [ ] `pnpm test:react-compat`
- [ ] `npm run ciTest`
- [x] Validation previously performed on the original branch version:
  - `pnpm lint`
  - `pnpm --filter @webspatial/core-sdk lint`
  - `pnpm --filter @webspatial/react-sdk lint`
  - `pnpm exec lint-staged --debug`
  - `prettier --check` on edited files
- [ ] Re-run lint and pre-commit timing validation after review fixes.

## Package release

- [ ] This PR does not change published packages under `packages/`.
- [x] I added a `.changeset/*.md` file.
- [ ] A maintainer should apply `skip-changeset`.

> `.changeset/eslint-precommit-governance.md` marks `@webspatial/core-sdk` and
> `@webspatial/react-sdk` as `patch` for internal tooling changes. The fixed SDK
> group may cause additional packages to appear in Changesets output.

## Screenshots or recordings

N/A — no UI or runtime behavior changes.

## Checklist

- [x] I kept this PR focused and avoided unrelated refactors.
- [x] I preserved existing inline comments unless changing them was part of this PR.
- [x] I updated documentation when the implementation changed.


openspec/changes/eslint-precommit-governance/specs/lint-governance/spec.md
```diff
@@ -0,0 +1,59 @@
+# Lint governance (WebSpatial SDK monorepo)
```
khufu-dev: I don't think this needs a spec. It's a bit overkill.

package.json
```diff
@@ -39,6 +40,26 @@
...
+      "prettier --write"
+    ],
+    "packages/core/src/**/*.{ts,tsx}": [
+      "eslint --fix",
+      "prettier --write"
+    ],
+    "packages/react/src/**/*.{ts,tsx}": [
+      "eslint --fix",
+      "prettier --write"
+    ],
```
khufu-dev: <sub><sub>P1 Badge</sub></sub> How long does this take? If it's more than 300ms I prefer that it's not in the pre-commit hook and instead be part of the build step.

eslint.config.js
```diff
@@ -0,0 +1,142 @@
...
+      },
+    },
+  },
+  {
+    files: [
+      'apps/test-server/**/*.{ts,tsx}',
+      'tests/ci-test/src/**/*.{ts,tsx}',
+    ],
+    languageOptions: {
+      ecmaVersion: 2020,
```
khufu-dev: This should be newer since we're developing new web technologies that are part of future standards.

eslint.config.js
```diff
@@ -0,0 +1,142 @@
...
+      '@typescript-eslint/no-unused-vars': 'off',
+      '@typescript-eslint/ban-ts-comment': 'off',
+      '@typescript-eslint/no-explicit-any': 'off',
+      '@typescript-eslint/no-require-imports': 'off',
+      '@typescript-eslint/no-unsafe-function-type': 'off',
+      '@typescript-eslint/no-wrapper-object-types': 'off',
+      'no-var': 'off',
+      'prefer-const': 'off',
+      'unused-imports/no-unused-imports': 'error',
+      'unused-imports/no-unused-vars': 'off',
```
khufu-dev: Will you enable these stricter lint rules in the future?

khufu-dev (changes requested):

openspec/changes/eslint-precommit-governance/specs/lint-governance/spec.md
mzvast: Agreed. I removed the OpenSpec change entirely (`proposal.md`, `design.md`, `tasks.md`, the spec, and `.openspec.yaml`) and kept the rationale in the PR description instead.

eslint.config.js
mzvast: Updated this to `ecmaVersion: 'latest'` so ESLint does not unnecessarily reject newer JavaScript syntax. Runtime/browser compatibility remains governed by the build toolchain.

eslint.config.js
mzvast: Rather than enabling the full recommended rule sets and disabling many rules, I simplified the shared config to explicitly enforce only `unused-imports/no-unused-imports`. Stricter rules can be introduced separately when the repository is ready for the associated cleanup.

package.json
mzvast: I optimized the pre-commit path by consolidating the ESLint tasks and enabling cache.

I don’t think 300 ms should be a hard cutoff. Projects such as Vite also run `eslint --cache --fix` on staged files. Doing this before commit avoids an unnecessary CI round trip for issues that can be fixed automatically.

The usual CI pattern is to check formatting and linting, not to modify the branch and commit fixes on behalf of the developer. If CI finds an issue, the developer normally fixes it locally and pushes again. Keeping fast, safe auto-fixes in pre-commit avoids that extra cycle.

khufu-dev: My concern with adding `eslint` as a pre-commit hook is that as our project grows it will become slower and block committing and saving code. Yes 300ms was a suggestion and definitely not a hard cutoff. The other challenge with making it a pre-commit hook is that eslint cannot automatically fix all problems and when there's an error the Git client may not show the error messages.
Can we start by adding it to the build or test command and later we can include it as a pre-commit hook if it becomes a problem? There are also faster options such as https://oxc.rs/docs/guide/usage/linter

PR #1318, opened 2026-07-09
fix: ModelLoadEvent typing to cover bubbled child load/error events

### LINKED ISSUE #1293: Model incorporate child load and error events in typings
Model can have child elements such as `` or `<video>` elements which trigger load and error events which bubble up to the Model. Currently the `ModelLoadEvent` type doesn't correctly reflect this runtime behaviour. 

Additionally the `target` and `currentTarget` should both be `SpatializedStatic3DElementRef` when triggered on the Model element. When triggered on a child element and the event listener is on `<Model>` then `currentTarget` should be portal object of type SpatializedStatic3DElementRef

```jsx
<Model onLoad={e => console.log(e)} onError={e => console.error(e)}>
  {/* <img> triggers load event on parent when source loads */}
  <img src="model_fallback.png" />
</Model>
```

khufu-dev:
ModelLoadEvent hard-typed target as the Model ref and omitted currentTarget. This breaks when a child /<video> fires a load/error that bubbles to the Model, target is the child element. Co-Authored-By Claude Opus 4.8

Fixes #1293

### DIFF
--- a/.changeset/model-load-event-child-target.md
+++ b/.changeset/model-load-event-child-target.md
@@ -1,0 +1,5 @@
+---
+'@webspatial/react-sdk': patch
+---
+
+Fix `ModelLoadEvent` typings to cover bubbled child load/error events (#1293).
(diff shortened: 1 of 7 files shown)


khufu-dev:
@codex review please

JackHFK (approved):

PR #1319, opened 2026-07-09
proposal for useEntityAnimation API

mzvast:
## Summary

This PR adds the OpenSpec proposal, design, specs, task list, and design-review record for redesigning the Entity motion API around `useEntityAnimation`.

The proposal defines `useEntityAnimation` as the Entity-specific surface over the shared `useAnimation` motion family:

- return `[animation, api, entityProps]`
- keep Entity authoring in `position` / `rotation` / `scale`
- support percentage `timeline` in addition to `from` / `to`
- recommend `xr-animation` while keeping `animation` compatible
- use `entityProps` as the React outlet for native-confirmed committed transform state
- add patch-object-only `api.set(values)` for committed transform writes
- keep Entity motion transform-only and explicitly reject unsupported targets such as `opacity`
- document per-component ownership, terminal fill-forward behavior, classified errors, and runtime capability guidance via `supports('useAnimation')`

This proposal supersedes `add-entity-transform-animation` as the target-state Entity motion proposal, while keeping the legacy proposal as historical context until implementation validation and archive/supersede follow-up are complete.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Tests or tooling
- [ ] Chore or maintenance

## Validation

- [ ] `pnpm test`
- [ ] `pnpm test:react-compat`
- [ ] `npm run ciTest`
- [ ] Manual validation:
- [x] Not run / not applicable: OpenSpec proposal/design/spec documentation only; no runtime implementation changes in this PR.

## Package release

- [x] This PR does not change published packages under `packages/`.
- [ ] I added a `.changeset/*.md` file.
- [ ] A maintainer should apply `skip-changeset`.

## Screenshots or recordings

Not applicable. This PR only adds OpenSpec documentation artifacts.

## Checklist

- [x] I kept this PR focused and avoided unrelated refactors.
- [x] I preserved existing inline comments unless changing them was part of this PR.
- [x] I updated docs, demos, or tests when behavior changed.


openspec/changes/redesign-entity-motion-api/proposal.md
```diff
@@ -0,0 +1,510 @@
...
+})
+
+return (
+  <BoxEntity {...entityProps} animation={animation} />
+)
+```
+
+**After the animation completes**, `entityProps` updates to the complete end pose (`position`, `rotation`, and `scale`), and the object remains on that committed pose. While the current binding lifecycle remains healthy, this complete mirror owns the transform even while playback is inactive. Animation-object creation or same-target config-replacement handoff failure clears the mirror and terminates that lifecycle.
+
+**When it updates**: `entityProps` does not update every frame. It only updates at key moments: when playback starts, completes, stops, resets, finishes, when an `api.set` write succeeds, and when creation or handoff failure clears it.
```
khufu-dev, line 315: I understand the reason behind this decision and the technical limitations. Do you think user's may get confused that entityProps only update at certain points? Also if entityProps get spread in the wrong order will it cause problems? For example

```javascript
<BoxEntity {...entityProps} animation={animation} position={basePosition} />
```

openspec/changes/redesign-entity-motion-api/proposal.md
```diff
@@ -0,0 +1,510 @@
...
+| Animation-object creation or pose handoff fails | The current binding lifecycle terminates, `entityProps` clears, and static/base React props control |
+| The animation binding is removed | React props regain control |
+
+This matches visionOS / picoOS natively: the underlying runtime binds the whole transform. During active animation, configured components animate and the remaining components hold their baseline values. Native reports the complete committed transform at confirmed lifecycle points, and `entityProps` persists that complete value while the current binding lifecycle remains healthy.
+
+A few practical takeaways follow:
+
+- **While the animation is playing**, the entire transform is taken over by the animation, so writing any component via props or `api.set` has no effect; components not in the config are frozen at baseline.
+- **During animation**, the runtime binds the whole transform and rotation holds its baseline value.
+- **After the first confirmed state**, ordinary transform props remain static/base inputs while the binding is attached. Use `api.set` for inactive dynamic changes.
```
khufu-dev: This is going to be confusing to the user. Sometimes the user can use the props directly on the Entity and sometimes they have to use `api`. Maybe we should simplify the confusion and say that the props have no effect when the `animation` property is set and the user has to use `useEntityAnimation` to transform the entity.

openspec/changes/redesign-entity-motion-api/proposal.md
```diff
@@ -0,0 +1,510 @@
...
+| `idle` | `false` | `false` | `false` |
+| `running` | `true` | `false` | `false` |
+| `paused` | `false` | `true` | `false` |
+| `finished` | `false` | `false` | `true` |
+
+### Behavior in Each State
+
+| State | How to enter | Is `api.set` usable | Does `entityProps` update | Who controls the transform |
+|---|---|---|---|---|
+| **Initial** | Before the first confirmed value | ✅ Usable after native object creation | Filled at the first confirmation | Static/base React props |
```
khufu-dev: If `entityProps` are updated at this point can it control the transform instead of base React props?

khufu-dev (approved):

openspec/changes/redesign-entity-motion-api/proposal.md
mzvast: Yes. Once confirmed, `entityProps` contains the complete transform and becomes the effective inactive pose when spread last. Base props provide the initial/fallback pose, and `api.set` updates the same confirmed pose through `entityProps`.

openspec/changes/redesign-entity-motion-api/proposal.md
mzvast: We considered that, but it would let animation state and `api.set` change the inactive pose behind the declared Entity props. We prefer the Element animation model: Native controls only active playback, while the composed React props control the inactive pose.

PR #1333, opened 2026-07-28
Update AttachmentEntity API compatability matrix

parthpatelsj:
##Summary

Adds compatibility for attachment placement on new runtimes without changing the existing public API.

- Detects AttachmentEntity:placement support on newer runtimes.
- Converts existing attachment options into the new placement payload.
- Preserves the existing payload for older runtimes.
- Leaves the new entity-style API to #1330.

### DIFF
--- a/.changeset/attachment-entity-placement.md
+++ b/.changeset/attachment-entity-placement.md
@@ -1,0 +1,5 @@
+---
+'@webspatial/core-sdk': patch
+---
+
+Fix attachments failing to load on new picoOS runtimes (PicoWebApp/0.4.90 OTA0+) by sending the placement-shaped attachment payload they expect, detected via `supports('AttachmentEntity', ['placement'])`. No API change; visionOS and older runtimes are unaffected.
--- a/.changeset/attachment-entity-placement.md
+++ b/.changeset/attachment-entity-placement.md
@@ -2,4 +2,4 @@
 '@webspatial/core-sdk': patch
 ---
 
-Fix attachments failing to load on new picoOS runtimes (PicoWebApp/0.4.90 OTA0+) by sending the placement-shaped attachment payload they expect, detected via `supports('AttachmentEntity', ['placement'])`. No API change; visionOS and older runtimes are unaffected.
+Send placement-shaped attachment payloads when the runtime advertises AttachmentEntity placement support.
--- a/packages/core/src/JSBCommand.ts
+++ b/packages/core/src/JSBCommand.ts
@@ -714,8 +714,8 @@ export class InitializeAttachmentCommand extends JSBCommand {
         ownerViewId: p.ownerViewId,
       }
     }
-    // Placement runtimes (PicoWebApp/0.4.90 OTA0+, WSAppShell/1.8.0+) only
-    // decode the placement-shaped payload: Vec3 position and meter dimensions.
+    // Runtimes advertising this capability decode a Vec3 position and meter
+    // dimensions instead of the legacy tuple and point size.
     return {
       id: this.attachmentId,
       placementId: p.parentEntityId,
--- a/packages/core/src/JSBCommand.ts
+++ b/packages/core/src/JSBCommand.ts
@@ -672,6 +672,8 @@ export class AnimateTransformJSBCommand extends JSBCommand {
     return params
   }
 }
+// TODO(remove): temporary OTA0 compat shim — delete once #1330 lands and
+// attachment JSB always uses the placement-shaped payload.
 function usePlacementProtocol(): boolean {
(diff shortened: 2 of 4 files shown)


packages/core/src/JSBCommand.ts
```diff
@@ -675,12 +686,25 @@ export class InitializeAttachmentCommand extends JSBCommand {
     super()
   }
   protected getParams() {
+    const p = this.options
+    if (!supports('AttachmentEntity', ['placement'])) {
```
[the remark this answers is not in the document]
parthpatelsj: Fixed — the WS_SHELL_VERSION visionOS debug placeholder now uses the legacy attachment payload, and visionOS is no longer marked as supporting the placement protocol. Only PicoWebApp 0.4.90+ receives the new payload. Once https://github.com/webspatial/webspatial-sdk/pull/1330 is merged in, then this will work.

packages/core/src/JSBCommand.ts
```diff
@@ -694,10 +736,20 @@ export class UpdateAttachmentEntityCommand extends JSBCommand {
...
+        ...this.options,
+      }
+    }
+    const { position, size } = this.options
+    const params: Record<string, any> = { id: this.attachmentId }
+    if (position) params.position = toPlacementVec3(position)
+    if (size) {
+      params.width = pointToPhysical(size.width)
+      params.height = pointToPhysical(size.height)
     }
```
khufu-dev: This is very poorly typed. Properties are dynamically added to `params` and the type checker cannot help with errors.

packages/core/src/JSBCommand.ts
```diff
@@ -666,6 +672,29 @@ export class AnimateTransformJSBCommand extends JSBCommand {
     return params
   }
 }
+function usePlacementProtocol(): boolean {
```
khufu-dev, line 677: This should be a member function of `InitializeAttachmentCommand` since it's not used outside.

packages/core/src/JSBCommand.ts
```diff
@@ -666,6 +672,29 @@ export class AnimateTransformJSBCommand extends JSBCommand {
...
+  if (
+    rt.type === 'visionos' &&
+    rt.shellVersion === VISIONOS_DEBUG_SHELL_VERSION_PLACEHOLDER
+  ) {
+    return false
+  }
+  return supports('AttachmentEntity', ['placement'])
+}
+
+function toPlacementVec3(position?: [number, number, number]): {
```
khufu-dev, line 691: This should also be a member function of InitializeAttachmentCommand

packages/core/src/JSBCommand.ts
```diff
@@ -1,4 +1,10 @@
 import { getPlatform } from './platform-runtime'
+import { pointToPhysical } from './physicalMetrics'
+import {
+  getRuntime,
+  supports,
+  VISIONOS_DEBUG_SHELL_VERSION_PLACEHOLDER,
+} from './runtime/supports'
```
khufu-dev, line 7: I'm a bit hesitant about adding runtime version checks in JSB since it could be called at a high frequency but I think this ok for now.

.changeset/attachment-entity-placement.md
```diff
@@ -0,0 +1,5 @@
+---
+'@webspatial/core-sdk': patch
+---
+
+Fix attachments failing to load on new picoOS runtimes (PicoWebApp/0.4.90 OTA0+) by sending the placement-shaped attachment payload they expect, detected via `supports('AttachmentEntity', ['placement'])`. No API change; visionOS and older runtimes are unaffected.
```
khufu-dev: Patch changesets should be one line.

khufu-dev (approved):

mzvast (approved):

packages/core/src/JSBCommand.ts
parthpatelsj: Fair point — switched to a plain object with conditional spreads instead of mutating Record<string, any>.

packages/core/src/JSBCommand.ts
parthpatelsj: Both init and update use this, so I left it file-private. Making it a member of only InitializeAttachmentCommand means duplicating it or adding a base class — overkill for this temp shim. 

This whole goes away once #1330 lands.

packages/core/src/JSBCommand.ts
parthpatelsj: Same — shared by both, so kept file-private.

packages/core/src/JSBCommand.ts
parthpatelsj: Yeah fair. This is only a temporary compat shim for OTA0 until #1330 merges — so should be fine for now.

PR #1342, opened 2026-08-06
fix(visionos): preserve double-click for child webviews

### LINKED ISSUE #1316: Custom scheme `window.open("custom://...")` breaks `dblclick` synthesis in child `WKWebView`
## Summary

In a visionOS SwiftUI host that embeds `WKWebView`, a child `WKWebView` created by `window.open("webspatial://...")` and returned from `WKUIDelegate.webView(_:createWebViewWith:for:windowFeatures:)` does not reliably synthesize DOM `dblclick` events. React `onDoubleClick` is also missing in the same child document.

The same native child `WKWebView` creation, injection, and display path works when the initial child URL is a standard URL such as `about:blank`.

## Expected behavior

When a user double clicks or double taps a normal DOM button inside the child `WKWebView`, WebKit should emit the standard event sequence:

```text
pointerdown
pointerup
click detail=1
pointerdown
pointerup
click detail=2
dblclick detail=2
```

React synthetic `onDoubleClick` should also fire when the button is rendered through a React portal into that child window.

## Actual behavior

When the child window is opened with `window.open("webspatial://...")`:

- DOM `dblclick` is missing inside the child document.
- React `onDoubleClick` is missing when using a cross-window React portal.
- The same host path receives basic pointer/click input, but the second-click synthesis chain differs from the `about:blank` control.

When the child window is opened with `window.open("about:blank")`, the child document can synthesize `click detail=2` and `dblclick detail=2`.

## Reproduction project

This branch includes a standalone visionOS repro project:

```text
apps/double-click-probe-visionos/DoubleClickProbe.xcodeproj
```

Run it on an Apple Vision Pro simulator, select a demo from the sidebar, double click the root and child probe buttons, and inspect Xcode console logs prefixed with `[DoubleClickProbe]`.

## Demo matrix

| Demo | Initial child URL | Host shape | React | Observed / expected |
| --- | --- | --- | --- | --- |
| demo0 | `about:blank` | direct root + SwiftUI ornament | no | `dblclick` works |
| demo1 | `about:blank` | `createWebViewWith` + ornament | no | `dblclick` works |
| demo2 | `about:blank` | ornament + WebSpatial-like body style | no | `dblclick` works |
| demo3 | `webspatial://createOrnament?...` | ornament + WebSpatial-like body style | no | local confirmation pending; expected to miss `dblclick` |
| demo4 | `about:blank` | ornament + head/style/class sync | no | `dblclick` works |
| demo5 | `about:blank` | ornament + React portal | yes | `dblclick` works |
| demo6 | `webspatial://createOrnament?...` | ornament + React portal | yes | `dblclick` is missing |
| demo7 | `about:blank` | nested child view + z offset | no | `dblclick` works |
| demo8 | `webspatial://createSpatialized2DElement?...` | nested child view + z offset | no | `dblclick` is missing |

## Why this points to the initial URL scheme

The control demos rule out the following variables:

- `WKWebView` itself lacking `dblclick` support.
- SwiftUI `.ornament`.
- `window.open` and `createWebViewWith` by themselves.
- WebSpatial-like viewport or body styles.
- head/style/link/className sync.
- React portal synthetic event handling.
- plain nested SwiftUI child view layout.
- z offset / depth placement.

The remaining distinguishing variable is the initial URL passed to `window.open`. The `about:blank` path works, while the `webspatial://...` path breaks child-window double-click synthesis.

## Possible fix direction

Avoid using `webspatial://...` as the real initial URL for the child browsing context.

Keep the existing synchronous `window.open` creation model, but move the spatial command data from the custom scheme/host into the query string of a standard URL:

```text
window.open("about:blank?command=createOrnament&attachmentAnchor=bottom&...", target, features)
```

In other words, avoid putting the command in the URL scheme or host:

```text
webspatial://createOrnament?...
```

Instead, parameterize the same data on `about:blank`:

```text
about:blank?command=createOrnament&attachmentAnchor=bottom&contentAlignment=back&visibility=visible&width=360&height=240
```

One possible flow:

1. JavaScript synchronously calls `window.open("about:blank?<encoded-webspatial-params>", target, features)`.
2. Native reads `navigationAction.request.url` in `createWebViewWith`.
3. Native parses `command` and spatial element parameters from the URL query.
4. Native creates the corresponding spatial container and returns the child `WKWebView`.
5. The child document continues to be written by the SDK or rendered through React Portal.

This preserves the original synchronous `window.open` creation model while avoiding a custom-scheme URL that can affect WebKit child-window event synthesis.

Test demo: [double-click-webspatial-window-open](https://github.com/webspatial/webspatial-sdk/tree/repro/double-click-webspatial-window-open/apps/double-click-probe-visionos)

JackHFK:
## Summary

Fixes #1316.

Fix visionOS child `WKWebView` double-click synthesis by avoiding `webspatial://...` as the initial `window.open` URL for spatial child windows.

## Problem

On visionOS, WebSpatial creates child windows for `SpatialDiv`, `Attachment`, and `Ornament` through:

```ts
window.open("webspatial://<command>?...")</command>
```

Native handles the request in `WKUIDelegate.webView(_:createWebViewWith:for:windowFeatures:)`, creates the corresponding native container, and returns a child `WKWebView`.

The user-facing issue was first observed on Ornament: DOM `dblclick` and React `onDoubleClick` inside an Ornament child `WKWebView` did not fire reliably. Further isolation showed the same root cause also applies to other spatial child windows created through the same custom-scheme `window.open` path, including `SpatialDiv`.

The repro matrix showed:

- `about:blank` child `WKWebView`: `click detail=2`, DOM `dblclick`, and React `onDoubleClick` work.
- `webspatial://createOrnament?...` child `WKWebView`: `dblclick` / React `onDoubleClick` do not fire reliably.
- `webspatial://createSpatialized2DElement?...` child `WKWebView`: the same double-click synthesis issue appears outside Ornament as well.
- React Portal, SwiftUI Ornament, viewport/style sync, nested views, and z offset were ruled out.

So the key variable is the initial URL scheme used by `window.open`.

## Fix

Instead of opening spatial child windows with the internal custom scheme on visionOS, the SDK now rewrites supported spatial content commands to a standard blank-page URL:

```text
about:blank?command=<command>&rid=<rid>&wsepoch=<wsepoch>&...</wsepoch></rid></command>
```

This keeps the synchronous `window.open -> WindowProxy` contract while avoiding the WebKit custom-scheme child browsing-context behavior that breaks double-click synthesis.

### Frontend

In `scene-polyfill.ts`:

- Keep the existing Pico token URL path unchanged when `genToken()` is available.
- For tokenless visionOS runtimes, rewrite these commands to `about:blank?...`:
  - `createSpatialized2DElement`
  - `createAttachment`
  - `createOrnament`
- Preserve request metadata and existing command params in the query string.
- Avoid treating all tokenless runtimes as visionOS by using the existing adapter resolver.

### Native visionOS

In `SpatialScene.swift`:

- Add unified command parsing for both:
  - legacy `webspatial://<command>?...`
  - new `about:blank?command=<command>&...`
- Keep `createSpatialScene` restricted to the legacy `webspatial://` path.
- Allow only spatial child content commands through `about:blank`.
- Parse `rid`, `wsepoch`, and Ornament create options from the same query source.
- Handle WebKit’s opaque `about:` URL behavior where `absoluteString` contains the query, but `url.query` / direct `URLComponents(string:)` parsing may not expose it.

In `SpatialWebController.swift`:

- Keep the original navigation decision order.
- Fold internal `about:blank?...` child navigation allowance into the existing denied-navigation fallback.
- Continue canceling internal `webspatial://` navigation and externally opening ordinary non-internal URLs.

## Demo Update

The Ornament test-server demo now includes a visible double-click probe inside Ornament content:

- `Double-click me` button
- visible counter
- stable test ids:
  - `ornament-dblclick`
  - `ornament-dblclick-count`

This makes manual verification of React `onDoubleClick` inside the child `WKWebView` straightforward.

## Validation

- `@webspatial/core-sdk` type check passed.
- Focused `scene-polyfill` request metadata tests passed.
- `apps/test-server` TypeScript check passed.
- SwiftFormat passed.
- visionOS Simulator build passed.
- Runtime verification on visionOS Simulator confirmed:
  - SpatialDiv child `WKWebView` creation returns non-null `WindowProxy`.
  - Ornament child `WKWebView` creation returns non-null `WindowProxy`.
  - Native command parsing, `rid`, `wsepoch`, and child `WKWebView` creation all succeed with `about:blank?...`.
  - Ornament demo renders the double-click test button.

### DIFF
--- a/.changeset/collapsible-sidebar-groups.md
+++ b/.changeset/collapsible-sidebar-groups.md
@@ -1,5 +1,0 @@
----
-'web-content': patch
----
-
-Make the test-server sidebar groups collapsible while keeping the active route group expanded.
--- a/.changeset/collapsible-sidebar-groups.md
+++ b/.changeset/collapsible-sidebar-groups.md
@@ -1,0 +1,5 @@
+---
+'web-content': patch
+---
+
+Make the test-server sidebar groups collapsible while keeping the active route group expanded.
--- a/.changeset/fix-visionos-child-window-dblclick.md
+++ b/.changeset/fix-visionos-child-window-dblclick.md
@@ -1,0 +1,8 @@
+---
+'@webspatial/core-sdk': patch
+'@webspatial/platform-visionos': patch
+---
+
+Fix visionOS child windows created through `window.open` so DOM double-click events continue to be synthesized.
+
+Core now rewrites tokenless visionOS spatial child-window commands to `about:blank` URLs with query parameters. The visionOS runtime accepts and parses those opaque `about:` URLs while preserving legacy `webspatial://` command support.
--- a/apps/test-server/src/components/Sidebar.tsx
+++ b/apps/test-server/src/components/Sidebar.tsx
@@ -1,3 +1,4 @@
+import { useEffect, useState } from 'react'
 import { Link, useLocation } from 'react-router-dom'
 import { entityAnimationRoutes } from '../pages/entity-animation/routes'
 import { spatialElementMotionRoutes } from '../pages/spatial-element-motion/routes'
@@ -78,13 +79,13 @@ export const routes = [
     ],
   },
   { path: '/unit-convert', label: 'Unit Convert' },
-  { path: '/scene/ornament-test', label: 'Ornament Test' },
   {
     path: '/scene',
     label: 'Scene',
     children: [
       { path: '/scene', label: 'Scene Landing' },
       { path: '/scene/volume', label: 'Volume' },
+      { path: '/scene/ornament-test', label: 'Ornament Test' },
       { path: '/scene/xrapp', label: 'XR App' },
       { path: '/scene/nosdk', label: 'No SDK' },
     ],
@@ -159,6 +160,41 @@ export default function Sidebar() {
         typeof child.path === 'string' &&
         location.pathname === child.path,
     )
+  const [expandedRoutes, setExpandedRoutes] = useState<Set<string>>(
+    () => new Set(items.filter(isRouteActive).map(route => route.path)),
+  )
+
+  useEffect(() => {
+    const activeParents = items
+      .filter(route => route.children && isRouteActive(route))
+      .map(route => route.path)
+
+    if (activeParents.length === 0) return
+
+    setExpandedRoutes(prev => {
+      const next = new Set(prev)
+      let changed = false
+      activeParents.forEach(path => {
+        if (!next.has(path)) {
+          next.add(path)
+          changed = true
+        }
+      })
+      return changed ? next : prev
+    })
+  }, [items, location.pathname])
+
+  const toggleRoute = (path: string) => {
+    setExpandedRoutes(prev => {
+      const next = new Set(prev)
+      if (next.has(path)) {
+        next.delete(path)
+      } else {
+        next.add(path)
+      }
+      return next
+    })
+  }
 
   return (
     <div className={containerClass}>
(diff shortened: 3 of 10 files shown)


JackHFK:
@codex review

parthpatelsj:
Thanks for this PR @JackHFK 

Wondering if this is still a workaround for a visionOS/WebKit event bug? There is no Apple or web-platform documentation stating that a child created from a custom URL scheme should lose click detail=2 or dblclick (unless you are aware of one)

I think we should file a standalone WebKit Bugzilla report and Apple Feedback using the isolated repro, then link them from the code/PR. https://bugs.webkit.org/

mzvast (commented):
Blocking compatibility concern: the new `about:blank?command=...` path is currently gated only by `resolveJsbAdapterPlatform(...) === 'visionos'`, not by the native shell capability that actually understands this protocol.

This is fine in local development where JS and native are built in lockstep, but after deployment the web/JS side can update before the user's installed app. In that case, a newer JS SDK could start sending `about:blank` commands to an older WSAppShell that only handles the legacy `webspatial://` path

mzvast (approved):

PR #1344, opened 2026-08-06
Support Attachment corner radius and background material

parthpatelsj:
## Summary

Adds support for WebSpatial Attachment `cornerRadius` and `backgroundMaterial`.

## Changes

* Parse both properties from Attachment initialize and update JSB commands.
* Store the properties in the spatial scene Attachment model.
* Apply corner radius and background material during Attachment creation and updates.
* Update the WebCore dependency required by the new Attachment API.
* Preserve existing Attachment behavior when the properties are omitted.

## Root cause

Attachment commands and runtime models did not propagate corner radius or background material to native rendering.

### DIFF
--- a/packages/core/src/JSBCommand.ts
+++ b/packages/core/src/JSBCommand.ts
@@ -28,6 +28,10 @@ import {
   SpatialTextureResourceOptions,
 } from './types/types'
 import type { OrnamentOptions } from './Ornament'
+import {
+  normalizeAttachmentBackgroundMaterial,
+  normalizeAttachmentCornerRadius,
+} from './reality/attachmentSurface'
 import type { AnimateTransformCommand } from './types/animation'
 import { composeSRT } from './utils'
 import type {
@@ -717,6 +721,10 @@ export class InitializeAttachmentCommand extends JSBCommand {
       width: this.options.width,
       height: this.options.height,
       ownerViewId: this.options.ownerViewId,
+      cornerRadius: normalizeAttachmentCornerRadius(this.options.cornerRadius),
+      backgroundMaterial: normalizeAttachmentBackgroundMaterial(
+        this.options.backgroundMaterial,
+      ),
     }
   }
 }
@@ -730,10 +738,23 @@ export class UpdateAttachmentEntityCommand extends JSBCommand {
     super()
   }
   protected getParams() {
-    return {
+    // Omitted fields stay omitted so the native side preserves the
+    // attachment's existing effective values on partial updates.
+    const params: Record<string, unknown> = {
       id: this.attachmentId,
       ...this.options,
     }
+    if (this.options.cornerRadius !== undefined) {
+      params.cornerRadius = normalizeAttachmentCornerRadius(
+        this.options.cornerRadius,
+      )
+    }
+    if (this.options.backgroundMaterial !== undefined) {
+      params.backgroundMaterial = normalizeAttachmentBackgroundMaterial(
+        this.options.backgroundMaterial,
+      )
+    }
+    return params
   }
 }
 
--- a/packages/core/src/JSBCommand.ts
+++ b/packages/core/src/JSBCommand.ts
@@ -23,6 +23,8 @@ import {
   Vec3,
   AttachmentEntityOptions,
   AttachmentEntityUpdateOptions,
+  BackgroundMaterialType,
+  CornerRadius,
   ModelLoadingMode,
   ModelSource,
   SpatialTextureResourceOptions,
@@ -740,19 +742,24 @@ export class UpdateAttachmentEntityCommand extends JSBCommand {
   protected getParams() {
     // Omitted fields stay omitted so the native side preserves the
     // attachment's existing effective values on partial updates.
-    const params: Record<string, unknown> = {
+    const { cornerRadius, backgroundMaterial, ...rest } = this.options
+    const params: {
+      id: string
+      cornerRadius?: CornerRadius
+      backgroundMaterial?: BackgroundMaterialType
+    } & Omit<
+      AttachmentEntityUpdateOptions,
+      'cornerRadius' | 'backgroundMaterial'
+    > = {
       id: this.attachmentId,
-      ...this.options,
+      ...rest,
     }
-    if (this.options.cornerRadius !== undefined) {
-      params.cornerRadius = normalizeAttachmentCornerRadius(
-        this.options.cornerRadius,
-      )
+    if (cornerRadius !== undefined) {
+      params.cornerRadius = normalizeAttachmentCornerRadius(cornerRadius)
     }
-    if (this.options.backgroundMaterial !== undefined) {
-      params.backgroundMaterial = normalizeAttachmentBackgroundMaterial(
-        this.options.backgroundMaterial,
-      )
+    if (backgroundMaterial !== undefined) {
+      params.backgroundMaterial =
+        normalizeAttachmentBackgroundMaterial(backgroundMaterial)
     }
     return params
   }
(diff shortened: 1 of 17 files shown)


packages/core/src/types/types.ts
```diff
@@ -433,6 +433,16 @@ export interface AttachmentEntityOptions {
   scale?: Vec3
   width?: number
   height?: number
+  /**
+   * Uniform corner radius in points applied to all four corners of the
+   * attachment surface. Invalid values (negative or non-finite) fall back to 0.
+   */
```
khufu-dev: AI likes to add big comments to everything but I think it's unnecessary.

packages/core/src/reality/attachmentSurface.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const DEFAULT_ATTACHMENT_BACKGROUND_MATERIAL: BackgroundMaterialType =
+  'transparent'
+
+/**
+ * Expands the public single-number attachment corner radius into the internal
+ * four-corner CornerRadius command shape. Missing, negative, or non-finite
+ * values fall back to 0 for all corners.
+ */
+export function normalizeAttachmentCornerRadius(value: unknown): CornerRadius {
```
khufu-dev: Why is it unknown, isn't a number?

```suggestion
export function normalizeAttachmentCornerRadius(value: number): CornerRadius {
```

packages/core/src/reality/attachmentSurface.ts
```diff
@@ -0,0 +1,43 @@
...
+    topTrailing: radius,
+    bottomTrailing: radius,
+  }
+}
+
+/**
+ * Missing or invalid materials fall back to 'transparent'.
+ */
+export function normalizeAttachmentBackgroundMaterial(
+  value: unknown,
```
khufu-dev: How come this is `unknown` type?

```suggestion
  value: BackgroundMaterialType,
```

packages/core/src/JSBCommand.ts
```diff
@@ -730,10 +738,23 @@ export class UpdateAttachmentEntityCommand extends JSBCommand {
     super()
   }
   protected getParams() {
-    return {
+    // Omitted fields stay omitted so the native side preserves the
+    // attachment's existing effective values on partial updates.
+    const params: Record<string, unknown> = {
```
khufu-dev: Can this be typed better so that the return type is more strictly typed

packages/react/src/reality/components/AttachmentEntity.tsx
```diff
@@ -1,20 +1,55 @@
...
-import type { Vec3 } from '@webspatial/core-sdk'
+import type { BackgroundMaterialType, Vec3 } from '@webspatial/core-sdk'
 
 import { useRealityContext, useParentContext } from '../context'
 import { setOpenWindowStyle } from '../../utils/windowStyleSync'
 import { useSyncHeadStyles } from '../../utils/useSyncHeadStyles'
 
 let instanceCounter = 0
 
+const ATTACHMENT_BACKGROUND_MATERIALS = new Set<string>([
```
khufu-dev: String is implied.

```suggestion
const ATTACHMENT_BACKGROUND_MATERIALS = new Set([
```

packages/react/src/reality/components/AttachmentEntity.tsx
```diff
@@ -1,20 +1,55 @@
...
+  'transparent',
+  'translucent',
+  'thin',
+  'regular',
+  'thick',
+])
+
+function normalizeCornerRadiusProp(value: number | undefined) {
+  if (value === undefined) return undefined
+  return typeof value === 'number' && Number.isFinite(value) && value >= 0
```
khufu-dev: This would convert infinity to 0?

packages/react/src/reality/components/AttachmentEntity.tsx
```diff
@@ -1,20 +1,55 @@
...
+  if (typeof value === 'string' && ATTACHMENT_BACKGROUND_MATERIALS.has(value)) {
+    return value
+  }
+  if (typeof process === 'undefined' || process.env.NODE_ENV !== 'production') {
+    console.warn(
+      `[AttachmentEntity] Invalid backgroundMaterial "${String(value)}"; ` +
+        `falling back to "transparent". Supported values: ` +
+        `'none' | 'transparent' | 'translucent' | 'thin' | 'regular' | 'thick'.`,
+    )
+  }
```
khufu-dev: This should be solvable via proper typescript types. Not a fan of runtime warnings if it can be identified at build.

packages/visionOS/web-spatial/manager/AttachmentManager.swift
```diff
@@ -8,11 +8,36 @@ struct AttachmentInfo: Identifiable, Equatable {
...
+        return CornerRadius(
+            topLeading: clampRadiusValue(value.topLeading),
+            bottomLeading: clampRadiusValue(value.bottomLeading),
+            topTrailing: clampRadiusValue(value.topTrailing),
+            bottomTrailing: clampRadiusValue(value.bottomTrailing)
+        )
+    }
+
+    static func clampRadiusValue(_ value: CGFloat) -> CGFloat {
+        value.isFinite && value >= 0 ? value : 0
```
khufu-dev, line 30: Didn't you already do this on the front end?

packages/visionOS/web-spatial/view/SpatializedDynamic3DView.swift
```diff
@@ -173,6 +173,11 @@ struct SpatializedDynamic3DView: View {
             ForEach(Array(spatialScene.attachmentManager.attachments.values)) { info in
                 Attachment(id: info.id) {
                     info.webViewModel.getView()
+                        .materialWithBorderCorner(
+                            info.backgroundMaterial,
+                            info.cornerRadius,
```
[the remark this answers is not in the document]
khufu-dev: This seems like a problem

khufu-dev (approved):

packages/react/src/reality/components/AttachmentEntity.tsx
```diff
@@ -136,10 +181,19 @@ export const AttachmentEntity: React.FC<AttachmentEntityProps> = ({
...
     if (!attachmentRef.current) return
-    attachmentRef.current.update({ position, rotation, scale, width, height })
+    attachmentRef.current.update({
+      position,
+      rotation,
+      scale,
+      width,
+      height,
+      cornerRadius: normalizedCornerRadius,
+      backgroundMaterial: normalizedBackgroundMaterial,
```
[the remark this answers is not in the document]
parthpatelsj: I don’t think this is an issue in the current flow because cornerRadius and backgroundMaterial are treated as initialization values and aren’t expected to change while the attachment is being created. I’d prefer not to add extra synchronization logic without a reproducible case, but canrevisit if there’s a supported path where these values can change during creation.

packages/visionOS/web-spatial/view/SpatializedDynamic3DView.swift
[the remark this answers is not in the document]
parthpatelsj: the glass clip comment was checked on device and translucent matches thick, this is a non-issue
Screenshot_2026-08-19_at_2 06 14_PM
Screenshot 2026-08-19 at 2 03 10 PM

packages/react/src/reality/components/AttachmentEntity.tsx
parthpatelsj: Yes, this is so native never gets a non-finite radius.

packages/react/src/reality/components/AttachmentEntity.tsx
parthpatelsj: Removed the React runtime console.warn, fair point.

packages/core/src/types/types.ts
parthpatelsj: I have removed it, to keep it clean and consistent with rest of the codebase

PR #1355, opened 2026-08-13
feat: add PBR material to dynamic 3d api

parthpatelsj:
## Summary

This PR adds physically based rendering (PBR) materials to the Reality material system alongside the existing `UnlitMaterial`.

PBR materials respond to scene lighting and can create matte, glossy, metallic, or glass-like surfaces. Their `roughness` and `metalness` properties follow the same naming and conventions as Three.js `MeshStandardMaterial`.

This first phase supports:

* `color`
* `textureId`
* `metalness`
* `roughness`
* `transparent`
* `opacity`

Emissive properties will be added in a follow-up.

## Core SDK

* Adds `SpatialSession.createPBRMaterial()`.
* Adds `SpatialPBRMaterial`, extending the abstract `SpatialMaterial` base.
* Adds the `CreatePBRMaterial` and `UpdatePBRMaterialProperties` bridge commands.
* Adds the `'pbr'` material sub-token.
* Adds the `PBRMaterial` runtime capability key.

## React SDK

* Adds a `<PBRMaterial>` component.
* Also supports PBR materials through `<Material type="pbr">` using a discriminated union.
* Exports the component through both the default lazy-loaded entry and the eager entry.
* All material props remain reactive after mount.
* Adds `MaterialPresets` with tuned starting points for `matte`, `glossy`, `plastic`, `metal`, and `glass` materials. These presets can be spread directly into `<PBRMaterial>`.
* Keeps `MaterialPresets` in a dependency-free top-level module so the default entry can export them without loading the spatial chunk.

This PR also refactors `UnlitMaterial` and `PBRMaterial` to share an internal `useSpatialMaterial` lifecycle hook. The hook handles:

* Native material creation and `ResourceRegistry` registration
* Texture ID resolution, including the existing tint-only fallback
* Late texture subscriptions
* Reactive property updates
* Cleanup when the component unmounts

The hook generalizes the existing `UnlitMaterial` lifecycle rather than replacing it, so `UnlitMaterial` behavior remains unchanged while both material types follow the same lifecycle.

## visionOS Native

* Adds `SpatialPBRMaterial`, backed by RealityKit’s `PhysicallyBasedMaterial`.
* Supports in-place property updates.
* Adds create and update handlers that mirror the existing unlit-material handlers.
* Extends texture-reload propagation to PBR materials.
* Renames `pushReloadedTextureToBoundUnlitMaterials` to `pushReloadedTextureToBoundMaterials` to reflect the broader behavior.

## Test Server

Adds a new `/reality/pbr-material` demo containing:

* One sphere for each material preset
* A configurable material demonstrating reactive updates to color, roughness, metalness, and transparency
* Coverage for the `<Material type="pbr">` dispatcher path


packages/core/src/runtime/keys.ts
```diff
@@ -21,6 +23,7 @@ export const COMPONENT_KEYS = [
   'ModelAsset',
   'ModelEntity',
   'UnlitMaterial',
+  'PBRMaterial',
```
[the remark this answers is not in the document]
walton007: Confirmed against the latest head (`74f53d7b`): this is still valid and should be fixed before merge. Because `baseTrueFlags()` derives every historical row from `TOP_LEVEL_KEYS` and `SUB_TOKENS_BY_NAME`, both `supports("PBRMaterial")` and `supports("Material", ["pbr"])` return `true` for old visionOS shells and the current Pico rows, even though those runtimes do not implement the new JSB commands. Please keep both flags false for existing rows/Pico and enable them only in the first visionOS shell row that actually ships the native handler; a regression test using an old `WSAppShell` UA and a current `PicoWebApp` UA would pin the contract.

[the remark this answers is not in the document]
parthpatelsj: Addressed in latest push

.changeset/pbr-material.md
```diff
@@ -0,0 +1,7 @@
+---
+'@webspatial/core-sdk': minor
+'@webspatial/react-sdk': minor
+'@webspatial/platform-visionos': minor
+---
+
+Add physically-based materials: `SpatialSession.createPBRMaterial()`, the `<PBRMaterial>` React component (also via `<Material type="pbr">`) with `MaterialPresets`, and RealityKit-backed rendering on visionOS.
```
khufu-dev: Make the description shorter.

packages/core/src/reality/material/SpatialMaterial.ts
```diff
@@ -10,5 +18,5 @@ export abstract class SpatialMaterial extends SpatialObject {
     this.type = type
   }
 
-  abstract updateProperties(properties: any): void
+  abstract updateProperties(properties: O): Promise<CommandResult>
```
khufu-dev: Thank you, stricter typing really helps our codebase.

packages/core/src/reality/material/SpatialMaterial.ts
```diff
@@ -1,7 +1,15 @@
...
+import type { SpatialMaterialType } from '../../types/types'
 
-export abstract class SpatialMaterial extends SpatialObject {
+/**
+ * Native material handle. `O` is the create/update property bag for that
+ * material kind (unlit, PBR, and any later type). Collections can use the
+ * default `SpatialMaterial` (i.e. `SpatialMaterial<object>`).
+ */
+export abstract class SpatialMaterial<
+  O extends object = object,
```
khufu-dev: `object` seems to broad for the type especially as default value. Is there a more restrictive type

walton007 (approved):

PR #1373, opened 2026-09-01
fix: Model entityTransform shouldn't affect poster

### LINKED ISSUE #1372: Model entityTransform should not affect poster
According to the [specs](https://immersive-web.github.io/model-element) and [model explainer](https://github.com/immersive-web/model-element/blob/main/explainer.md#stage-interaction-mode)  `entityTransform` should not affect the poster image, only the 3D model. CSS transform of the `<Model>` element does transform the poster. The behaviour is broken on visionOS but correct on PICO OS 6.

- OS: visionOS 26
- WebSpatial version: 2.0
- [Meego](https://meego.larkoffice.com/pico/issue/detail/7366844609)

khufu-dev:
According to the specs and model explainer  entityTransform should not affect the poster image, only the 3D model. CSS transform of the <Model> element does transform the poster.

Fixes #1372

https://github.com/user-attachments/assets/f56e7d0e-ced3-46f9-be02-a83ed43fa67b

### DIFF
--- a/packages/visionOS/web-spatial/view/SpatializedStatic3DView.swift
+++ b/packages/visionOS/web-spatial/view/SpatializedStatic3DView.swift
@@ -48,6 +48,10 @@ struct SpatializedStatic3DView: View {
                             .if(!depth.isZero) { view in view.scaledToFit3D() }
                             .if(enableGesture) { view in view.hoverEffect() }
                     }
+                    .scaleEffect(scale)
+                    .rotation3DEffect(rotation)
+                    .offset(x: x, y: y)
+                    .offset(z: z)
                 case .failed:
                     posterView {
                         // Transparent view is required so that lifecycle modifiers like .task still work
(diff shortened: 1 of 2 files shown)


parthpatelsj (approved):

packages/visionOS/web-spatial/view/SpatializedStatic3DView.swift
```diff
@@ -48,22 +48,22 @@ struct SpatializedStatic3DView: View {
...
                 case .failed:
                     posterView {
                         // Transparent view is required so that lifecycle modifiers like .task still work
                         Color.clear
                     }
                 }
             }
-            .scaleEffect(scale)
-            .rotation3DEffect(rotation)
             .orbit(
```
parthpatelsj, line 62: I assume leaving .orbit won't cause unintended interactions with the poster?

khufu-dev: It's currently disabled via `asset != nil`. Should I move it in for clarity?

Issue #1327, opened 2026-07-17
Model support blob URLs


khufu-dev:
## Summary
Add support for **blob URLs** as a source for the `<Model>` component. The blob URL should be accepted on the `src` attribute and on child `<source>` elements.

```jsx
const blobUrl = URL.createObjectURL(blob);
<Model enable-xr src={blobUrl} />
```

## Performance targets
These are the observed payload sizes the feature must handle
| Metric | Value |
|--------|-------|
| Size | 128 MB |
| Triangle count | 1.69 million |
| Vertex count | 1.01 million |
| Texture size | 8.6 MB |

[Technical Design](https://github.com/webspatial/webspatial-sdk/blob/main/docs/Model.md#7-blob-url-support-for-model-sources)
[Meego](https://meego.larkoffice.com/pico/story/detail/7358036034)

PR #1389, opened 2026-09-10
feat: Model support blob URLs Swift plumbing

khufu-dev:
Add support for **blob URLs** as a source for the `<Model>` component. The blob URL should be accepted on the `src` attribute and on child `<source>` elements.

```jsx
const blobUrl = URL.createObjectURL(blob);
<Model enable-xr src={blobUrl} />
```

[Technical Design](https://github.com/webspatial/webspatial-sdk/blob/main/docs/Model.md#7-blob-url-support-for-model-sources)
Issue #1327


packages/visionOS/web-spatial/JSBCommand.swift
```diff
@@ -439,3 +439,38 @@ struct UpdateAttachmentEntityCommand: CommandDataProtocol {
...
     let backgroundMaterial: BackgroundMaterial?
 }
+
+protocol BlobTransferCommand: CommandDataProtocol {
+    var id: String { get }
+    var requestId: String { get }
+}
+
+struct StartBlobTransfer: BlobTransferCommand {
+    static let commandType = "StartBlobTransfer"
```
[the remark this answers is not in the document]
khufu-dev: Yes it's in the next commit https://github.com/webspatial/webspatial-sdk/pull/1390

packages/visionOS/web-spatialTests/EntityMotionTests.swift
```diff
@@ -289,11 +289,11 @@ final class EntityMotionBridgeTypesTests: XCTestCase {
                 try JSONSerialization.jsonObject(
                     with: JSONEncoder().encode(decoded),
                     options: .fragmentsAllowed
-                ) as! NSObject,
+                ) as? NSObject,
```
parthpatelsj, line 292: nit: Doesn't this make the test fail when both casts fail i.e nil=nil? Maybe use `XCTUnwrap`

parthpatelsj (approved):

packages/visionOS/web-spatialTests/EntityMotionTests.swift
khufu-dev: This is a formatting change by `swiftformat` unrelated to this pull request .

PR #1397, opened 2026-09-14
fix: failing CodeQL Swift GitHub workflow

khufu-dev:
Xcode started emitting additional errors. The old script treated any error not matching exact message as a failure. It therefore exited the Build visionOS app step.


khufu-dev:
Incorrect fix

